PR #24, opened 2025-04-09
Hardened the  Rebalancer

andreimontchik:
1. Non functional changes to make the rebalancing processes more durable.
2. Some minor logging improvements.

### DIFF
--- a/src/rebalancer.rs
+++ b/src/rebalancer.rs
@@ -69,6 +69,7 @@ pub struct Rebalancer {
     mint_to_bank: HashMap<Pubkey, Pubkey>,
     oracle_to_bank: HashMap<Pubkey, Pubkey>,
     preferred_mints: HashSet<Pubkey>,
+    //FIXME: consider remove the Option wrapper. Rebalancing will not work w/o the swap mint bank.
     swap_mint_bank_pk: Option<Pubkey>,
     geyser_receiver: Receiver<GeyserUpdate>,
     stop_liquidations: Arc<AtomicBool>,
@@ -374,7 +375,7 @@ impl Rebalancer {
             || self.has_liabilities()
     }
 
-    fn rebalance_accounts(&mut self) -> anyhow::Result<()> {
+    fn fetch_swb_prices(&self) -> anyhow::Result<()> {
         let active_banks = self.liquidator_account.account_wrapper.get_active_banks();
 
         let active_swb_oracles: Vec<Pubkey> = active_banks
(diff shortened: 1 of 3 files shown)


src/rebalancer.rs
```diff
@@ -625,41 +640,48 @@ impl Rebalancer {
...
 
     fn deposit_preferred_tokens(&self) -> anyhow::Result<()> {
-        let balance = self.get_token_balance_for_bank(&self.swap_mint_bank_pk.unwrap())?;
-
-        if balance.is_none() {
-            return Ok(());
-        }
-
-        let balance = balance.unwrap();
+        if let Some(swap_mint_bank) = self.swap_mint_bank_pk {
```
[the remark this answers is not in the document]
IliaZyrin: Actually I intentionally put this early-return block here, to decrease nesting. Now you're bringing it back=) Why is that?
```
if balance.is_none() {
            return Ok(());
        }

        let balance = balance.unwrap();

```

src/rebalancer.rs
```diff
@@ -625,41 +640,48 @@ impl Rebalancer {
...
+                    );
 
-        if balance.is_zero() {
-            return Ok(());
+                    if let Some(bank) = self.banks.get(&swap_mint_bank) {
+                        if let Some(token_address) = self
+                            .token_account_manager
+                            .get_address_for_mint(bank.bank.mint)
+                        {
+                            if let Err(error) = self.liquidator_account.deposit(
```
IliaZyrin, line 656: Apart from nesting, this is now also not propagating the error back to the caller. I think we should propagate it.

src/rebalancer.rs
```diff
@@ -674,31 +696,34 @@ impl Rebalancer {
...
     fn drain_tokens_from_token_accounts(&mut self) -> anyhow::Result<()> {
-        let token_accounts: Vec<TokenAccountWrapper> =
-            self.token_accounts.values().cloned().collect();
-        for account in token_accounts {
-            if account.bank.bank.mint == self.config.swap_mint {
-                continue;
-            }
-
-            let value = account.get_value().unwrap();
+        if let Some(swap_mint_bank) = self.swap_mint_bank_pk {
```
IliaZyrin, line 707: I think this should stay as `unwrap()` given that if it's `None`, `rebalancer` is anyway fucked. I'm of the opinion that the app should `panic!` when there is a critical issue you have no handling for (and this is the case here, even though it can only happen if there is a programmer error, which would be caught right away, on first run, if we keep `unwrap()`)

IliaZyrin (changes requested):

src/rebalancer.rs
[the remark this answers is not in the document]
andreimontchik: Addressed on this morning's huddle.

src/rebalancer.rs
andreimontchik: Addressed on this morning's huddle.

IliaZyrin (approved):

PR #26, opened 2025-04-14
An improvement for building the remaining accounts list for liquidation

andreimontchik:
1. Streamlined the remaining accounts creation.
2. Parameterized the program CU limit.

### DIFF
--- a/src/wrappers/marginfi_account.rs
+++ b/src/wrappers/marginfi_account.rs
@@ -1,5 +1,6 @@
 use super::{bank::BankWrapperT, oracle::OracleWrapperTrait};
 use fixed::types::I80F48;
+use indexmap::IndexSet;
 use log::debug;
 use marginfi::state::marginfi_account::{BalanceSide, LendingAccount};
 use solana_program::pubkey::Pubkey;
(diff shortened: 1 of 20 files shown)


src/wrappers/marginfi_account.rs
```diff
@@ -111,35 +112,25 @@ impl MarginfiAccountWrapper {
...
-        &self,
+        lending_account: &LendingAccount,
         include_banks: &[Pubkey],
         exclude_banks: &[Pubkey],
         banks: &HashMap<Pubkey, BankWrapperT<T>>,
     ) -> Vec<Pubkey> {
-        // This is a temporary fix to ensure the proper order of the remaining accounts.
-        // It will NOT be necessary once this PR is deployed: https://github.com/mrgnlabs/marginfi-v2/pull/320
-        let active_bank_pks = self.get_active_banks();
+        let mut bank_pks: IndexSet<Pubkey> = IndexSet::new();
```
andreimontchik, line 120: IndexSet preserved input order and guarantees uniqueness. For free.

IliaZyrin (changes requested):

src/wrappers/marginfi_account.rs
```diff
@@ -111,35 +112,25 @@ impl MarginfiAccountWrapper {
     }
 
     pub fn get_observation_accounts<T: OracleWrapperTrait>(
-        &self,
+        lending_account: &LendingAccount,
```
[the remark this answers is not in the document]
andreimontchik: The only MarginfiAccountWrapper instance member that is used in the method is lending account.  Using static function in this case would be preferable IMHO, however this is largely matter of preference. 
The reason for the change was mostly to create unit tests for it, however after completing the change I found out that the method already has unit tests! :-) Updating tests as well and will add more in a bit. Hopefully the change will make more sense then.

andreimontchik:
Back to draft, the change exposed few more boo-boos.

src/wrappers/marginfi_account.rs
andreimontchik: Yeah, I jumped the gun with uniquness, misunterstood the program unit tests. Replacing it with lists.

I looked at the way the on-chain liquidation program obtains and loops thru active banks for liquidator and liquidatee accounts and could not figure out how squeezing active banks in place of inactive ones could help, in fact it probably breaks it instead. Also, I am not sure that include or exclude banks is applicable here at all. 🤷🏻 Disabled their usage for now. 
https://github.com/mrgnlabs/marginfi-v2/blob/3a4273d3a07694f99cc998eeaaad0618d95c09d9/programs/marginfi/src/state/marginfi_account.rs#L199

andreimontchik: Rolledbak the logic to the prior fix.

IliaZyrin: This is not quite right still. You removed this check which was important:
`if !active_bank_pks.contains(bank_pk)`

Without it, if both `active banks` and `include_banks` contain the same bank,  you may add it to the wrong place - first gap, whereas it should have been added later.

This might be a good test case btw:

Lending account balances: `[0: {active: true, bank: "a"}, 1: {active: false, bank:""}, 2: {active: false, bank: ""}, 3: {active: true, bank: "d"}]`
include_banks: `["d", "c"]`

Expected output:
`["a", "a_oracle", "c", "c_oracle", "d", "d_oracle"]`

whereas with your current logic it would return:
`["a", "a_oracle", "d", "d_oracle", "c", "c_oracle"]`

IliaZyrin (changes requested):

src/wrappers/marginfi_account.rs
andreimontchik: Thanks! Rolled it back. Adding more rather twisted logic does not make it any better than the original implementation.

PR #27, opened 2025-04-17
Refactoring the Solana clock fetching and usage

andreimontchik:
1. Refactored the Solana clock thread safe wrapper by separating clock data from the clock update process. Added the new solana clock refresh period parameter to the config toml file : `solana_clock_refresh_interval = 10`.  10 seconds for now.
2. Fixed the Rebalancer freezing bug! Root cause was blocking async Solana RPC client API, apparently it is trying to spin off another async call in current thread that is already blocked by waiting for it to complete! Deadlock at it's bestest! :-) Creating and using dedicated Tokio runtime with few threads solved the issue. 
3. Finally decoupled the liquidator's main() method from Tokio runtime. Instead the liquidator is now creating runtime instances with smaller thread pool sizes that are  incapsulated in processes that are using third party async API. This helps to better manage rather insane number of Tokio threads.


src/cli/entrypoints.rs
```diff
@@ -78,38 +93,43 @@ pub fn run_liquidator(config: Eva01Config) -> anyhow::Result<()> {
...
+        if let Err(e) = transaction_checker.start(jito_rx, pending_bundles) {
+            panic!("TransactionChecker failed! {:?}", e);
+        }
     });
 
-    tokio::task::spawn(async move {
+    thread::spawn(move || {
         if let Err(e) = rebalancer.start() {
-            panic!("Rebalancer failed: {:?}", e);
+            panic!("Rebalancer failed! {:?}", e);
```
jgur-psyops, line 127: would it make sense not to panic here, seems recoverable or even ignorable if the balancer is offline for some time. not sure really just wondering.

andreimontchik: No, not really, the Liquidator will be in bad state when Rebalancer goes down. It is better to shut it down and let Kubernetes to restart it. As bonus, we will have nice and scary errors logs as well as pagerduty alert. :-)

src/config.rs
```diff
@@ -166,6 +167,10 @@ impl GeneralConfig {
             pubkey!("5FuKF7C1tJji2mXZuJ14U9oDb37is5mmvYLf4KwojoF1"),
         ]
     }
+
+    pub fn default_sol_clock_refresh_interval() -> u64 {
+        10 // 10 seconds
```
jgur-psyops, line 172: so basically the system clock is up to 10s, or around 22 slots stale, and we no longer briefly cache the clock, we just use whatever the last fetched clock was? seems reasonable.

src/transaction_checker.rs
```diff
@@ -126,60 +134,63 @@ impl TransactionChecker {
...
+        Ok(())
     }
-}
 
-fn check_final_bundle_status(jito_sdk: &JitoJsonRpcSDK, uuid: &str) -> anyhow::Result<()> {
-    let max_retries = 10;
-    let retry_delay = std::time::Duration::from_millis(500);
+    fn check_final_bundle_status(&self, uuid: &str) -> anyhow::Result<()> {
+        let max_retries = 10;
+        let retry_delay = std::time::Duration::from_millis(500);
```
jgur-psyops, line 142: 10 * 500ms = 5s, is that sufficient?

src/config.rs
andreimontchik: Yep! This is temporary though. It is better to subscribe the Geyser client to the Solana clock account updates instead of constantly calling RPC. But this could wait, we will most likely largely redo the Liquidator while migrating it to our in-house data feeds instead of using RPC.

src/transaction_manager.rs
```diff
@@ -88,11 +89,17 @@ impl TransactionManager {
             lookup_tables.push(lookup_table);
         }
 
+        let tokio_rt = Builder::new_multi_thread()
+            .thread_name("transaction-manager")
+            .worker_threads(2)
```
jgur-psyops, line 94: why multithread the tx sender, the rpc will execute them serially regardless

src/transaction_checker.rs
andreimontchik: No idea honestly, I have not changed it.

jgur-psyops (approved):
lgtm, but I don't have nearly enough insight into this to assess thread safety.

src/transaction_manager.rs
andreimontchik: This is the price for not using async functions. I narrowed it down to tokio threads.

andreimontchik:
> lgtm, but I don't have nearly enough insight into this to assess thread safety.

Thanks! I am seriously doubtful anyone does. :-) Tokio threading api has became a real beast unfortunately.

src/transaction_checker.rs
andreimontchik: Five seconds total plus the RPC call time x 10. I am gonna double it in the next PR.

src/config.rs
IliaZyrin: Doesn't this lead to sometimes passing stale oracles to liquidation instructions?

src/transaction_checker.rs
IliaZyrin: In their [official example](https://github.com/jito-labs/jito-rust-rpc/blob/master/examples/basic_bundle.rs#L115) it's even more (20s for pre-status + 20s for the final one), but I just thought it's too much.

PR #28, opened 2025-04-18
Rebalancing improvements

andreimontchik:
1. Added the dust check liability repayments. 
2. Increased the time bundle retry time.
3. Some performance improvements.
4. Added logging for some basic metrics.

### DIFF
--- a/src/cli/entrypoints.rs
+++ b/src/cli/entrypoints.rs
@@ -149,14 +149,14 @@ pub fn run_liquidator(config: Eva01Config) -> anyhow::Result<()> {
     let monitor_transaction_rx = transaction_rx.clone();
     while !stop_liquidator.load(std::sync::atomic::Ordering::SeqCst) {
         thread_info!(
-            "Channel depth: [Lq,Rb,Tnx,Jito] -> [{},{},{},{}]",
+            "Channel depth: [Liquidation, Rebalancer, Txn, Jito] -> [{}, {}, {}, {}]",
             monitor_liquidator_rx.len(),
             monitor_rebalancer_rx.len(),
             monitor_transaction_rx.len(),
             monitor_jito_rx.len()
         );
         thread_info!(
-            "[Lq att, Failed Lqs] -> [{},{}]",
+            "Liqudations [attemps, failed] -> [{},{}]",
             LIQUIDATION_ATTEMPTS.get(),
             FAILED_LIQUIDATIONS.get()
         );
--- a/src/cli/entrypoints.rs
+++ b/src/cli/entrypoints.rs
@@ -149,14 +149,14 @@ pub fn run_liquidator(config: Eva01Config) -> anyhow::Result<()> {
     let monitor_transaction_rx = transaction_rx.clone();
     while !stop_liquidator.load(std::sync::atomic::Ordering::SeqCst) {
         thread_info!(
-            "Channel depth: [Liquidation, Rebalancer, Txn, Jito] -> [{}, {}, {}, {}]",
+            "Stats: Channel depth [Liquidation, Rebalancer, Txn, Jito] -> [{}, {}, {}, {}]",
             monitor_liquidator_rx.len(),
             monitor_rebalancer_rx.len(),
             monitor_transaction_rx.len(),
             monitor_jito_rx.len()
         );
         thread_info!(
-            "Liqudations [attemps, failed] -> [{},{}]",
+            "Stats: Liqudations [attemps, failed] -> [{},{}]",
             LIQUIDATION_ATTEMPTS.get(),
             FAILED_LIQUIDATIONS.get()
         );
--- a/src/cli/entrypoints.rs
+++ b/src/cli/entrypoints.rs
@@ -1,9 +1,11 @@
 use crate::{
     clock_manager::{self, ClockManager},
     config::Eva01Config,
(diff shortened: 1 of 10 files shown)


src/cli/entrypoints.rs
```diff
@@ -128,9 +136,33 @@ pub fn run_liquidator(config: Eva01Config) -> anyhow::Result<()> {
...
+
+    thread_info!("Entering main loop.");
+    let monitor_liquidator_rx = liquidator_rx.clone();
+    let monitor_rebalancer_rx = rebalancer_rx.clone();
+    let monitor_jito_rx = jito_rx.clone();
+    let monitor_transaction_rx = transaction_rx.clone();
+    while !stop_liquidator.load(std::sync::atomic::Ordering::SeqCst) {
+        thread_info!(
+            "Stats: Channel depth [Liquidation, Rebalancer, Txn, Jito] -> [{}, {}, {}, {}]",
+            monitor_liquidator_rx.len(),
```
IliaZyrin, line 153: We should probably add it to the Prometheus metrics, if it's not a temporary thing used for debugging.

src/cli/entrypoints.rs
```diff
@@ -92,6 +95,10 @@ pub fn run_liquidator(config: Eva01Config) -> anyhow::Result<()> {
     for (key, value) in rebalancer.get_accounts_to_track() {
         accounts_to_track.insert(key, value);
     }
+    accounts_to_track.insert(
```
IliaZyrin, line 98: Wasn't it there before?

src/marginfi_ixs.rs
```diff
@@ -21,13 +21,13 @@ pub fn make_deposit_ix(
...
-        marginfi_group,
         marginfi_account,
-        signer,
-        bank: bank.address,
+        authority: signer,
         signer_token_account,
-        bank_liquidity_vault: bank.bank.liquidity_vault,
+        liquidity_vault: bank.bank.liquidity_vault,
         token_program,
+        bank: bank.address,
```
IliaZyrin, line 29: Is reordering here - to match the order of members in the updated instruction?

src/wrappers/marginfi_account.rs
```diff
@@ -107,26 +109,29 @@ impl MarginfiAccountWrapper {
...
             .iter()
-            .filter_map(|balance| balance.active.then_some(balance.bank_pk))
+            .filter_map(|balance| balance.is_active().then_some(balance.bank_pk))
             .collect::<Vec<_>>()
     }
 
     // TODO: add more unit tests
     pub fn get_observation_accounts<T: OracleWrapperTrait>(
-        lending_account: &LendingAccount,
+        account_wrapper: &MarginfiAccountWrapper,
```
IliaZyrin, line 118: I was intentionally passing only the `lending_account` part here, so that the function doesn't get access to what it doesn't use. Plus it clearly shows the scope of the function to the caller. IMHO it's not worth it passing the whole `account_wrapper` for the sake of one debug log (it can be just moved up, to the caller level).

src/transaction_checker.rs
```diff
@@ -138,8 +138,8 @@ impl TransactionChecker {
     }
 
     fn check_final_bundle_status(&self, uuid: &str) -> anyhow::Result<()> {
-        let max_retries = 10;
-        let retry_delay = std::time::Duration::from_millis(500);
+        let max_retries = 5;
+        let retry_delay = std::time::Duration::from_secs(2);
```
IliaZyrin, line 142: I initially decreased the sleep time here to not delay next liquidations by too much, but I understand that it may be not enough to progress. We should probably just think of a way to trigger multiple liquidations at a time...

src/wrappers/liquidator_account.rs
```diff
@@ -166,12 +169,8 @@ impl LiquidatorAccount {
             })
             .collect::<Vec<_>>();
 
-        debug!(
-            "liquidate: observation_swb_oracles length: {:?}",
-            observation_swb_oracles.len()
-        );
-
         let crank_data = if !observation_swb_oracles.is_empty() {
+            thread_debug!("Cranking Swb Oracles for observation accounts.",);
```
IliaZyrin, line 173: IMHO, ending single-sentence logs with full stops doesn't convey any additional meaning since they're anyway separated by new lines from each other. But if it's important to you, we can keep it as-is, we will just have to make sure to use it consistently everywhere.

src/cli/entrypoints.rs
andreimontchik: Hm, not sure. Is it in the commit history?

src/marginfi_ixs.rs
andreimontchik: Correct.

src/wrappers/liquidator_account.rs
andreimontchik: The debug statement to confirm that Swb Oracle where attempted to be cranked, I did it for troubleshooting. We can clean up logging later.

src/wrappers/marginfi_account.rs
andreimontchik: Updated.

IliaZyrin (approved):

src/cli/entrypoints.rs
andreimontchik: I see. Alright,  I will remove the additional subscription I added. Thanks!

PR #29, opened 2025-04-28
Liquidation fixes

IliaZyrin:


src/cli/entrypoints.rs
```diff
@@ -113,33 +109,38 @@ pub fn run_liquidator(config: Eva01Config) -> anyhow::Result<()> {
 
     thread::spawn(move || {
         if let Err(e) = geyser_service.start() {
-            panic!("Geyser service failed! {:?}", e);
```
andreimontchik, line 116: What is the reason for this change? The idea is to shut down Eva in case of one of the processing threads failure and log an ugly and easily noticeable error.

IliaZyrin: Discussed on Slack.

src/cache/tokens.rs
```diff
@@ -32,7 +32,7 @@ impl TokensCache {
     ) -> anyhow::Result<()> {
         self.tokens
             .write()
-            .map_err(|e| anyhow!("Failed to lock the token map for insert! {}", e))?
+            .map_err(|e| anyhow!("Failed to lock the token map for insert: {}", e))?
```
andreimontchik, line 35: No bid deal, but the exclamation mark was on purpose here.

src/cache_loader.rs
```diff
@@ -512,6 +509,7 @@ impl CacheLoader {
                     &[&self.signer],
                     recent_blockhash,
                 );
+                info!("Sending transaction...");
```
andreimontchik, line 512: What transaction? The logging is not very helpful w/o some clarification.

src/cache/tokens.rs
IliaZyrin: I changed it everywhere where there was a chain of exclamation marks, without it it's harder to debug because you don't know where the "prefix" part starts and where is the actual error.
And in general IMHO it's not beneficial to have any exclamation marks at all, since it's already formatted as error anyway and all these symbols (and wordy phrases as well) are redundant.

src/cache_loader.rs
```diff
@@ -496,8 +495,6 @@ impl CacheLoader {
     }
 
     fn create_token_accounts(&self, instructions: Vec<Instruction>) -> Result<()> {
-        info!("Registering {} Token accounts...", instructions.len());
```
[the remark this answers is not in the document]
IliaZyrin: Because it's duplicating the log above (in the caller function).

src/cache_loader.rs
IliaZyrin: Ops! That was for debug. Will remove!

src/cache/tokens.rs
andreimontchik: Cool, thanks, I will follow the suit.

src/cache_loader.rs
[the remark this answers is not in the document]
andreimontchik: I see. Would you mind moving the concluding logging statement from the method as well? For consistency.

andreimontchik (approved):

PR #30, opened 2025-04-29
Implemented Cache

andreimontchik:
1. Implemented thread safe Cache for storing accounts, banks, oracles and tokens information.
2. Replaced process specific caches with the new Cache. 
3. Added the new GeyserProcessor service for processing and caching Geyser updates. The service sets liquidation and rebalancing flags if an update is triggering liquidation or rebalancing.
4. Hardened error handling in the liquidation and rebalancing processes.
5. Added unit tests.


LevBeta:
Having a built-in persistent cache in each liquidator instance is problematic, especially with the permissionless pools. As new pools can be created at any time, the cache can grow unbounded, even without that would be problematic with users increment on mrgnlend ,leading to excessive memory usage and potential VM crashes. It also increases the risk of stale data and creates redundant work across instances. A more scalable approach would be to use scoped, ephemeral caches or a shared external caching layer.

LevBeta:
Ephemeral caches were part of the original idea for a faster, stateless liquidator that avoids unnecessary overhead. The goal was to reduce duplicated work across a liquidator or cluster of liquidators, where each instance currently processes the same data independently—even when most of that work is redundant. By using short-lived, task-specific caches instead of a full persistent state, we can greatly improve performance, lower memory usage, and make the system more efficient overall.

andreimontchik:
> Ephemeral caches were part of the original idea for a faster, stateless liquidator that avoids unnecessary overhead. The goal was to reduce duplicated work across a liquidator or cluster of liquidators, where each instance currently processes the same data independently—even when most of that work is redundant. By using short-lived, task-specific caches instead of a full persistent state, we can greatly improve performance, lower memory usage, and make the system more efficient overall.

Hey hey! Long time no see! 🍻  Thanks for looking into this.

The cache is in fact ephemeral, not even sure why you think it is persistent. :-) Unlike the original design though it is shareable between processes. Reasoning for it is to first decouple data from processing and to avoid duplicate caches of the same data in different processes. This greatly simplifies maintenance and might slightly improve performance.
Another big problem this PR is fixing is the fact that rebalancing and liquidation are running in the same thread that is draining the Geyser messages channels. Since rebalancing and particularly liquidation is time consuming, the channels are growing in size. That is causing liquidations to fail due to outdated oracle and account information and eventually leads to OOM.

src/cli/entrypoints.rs
```diff
@@ -50,108 +54,134 @@ pub fn run_liquidator(config: Eva01Config) -> anyhow::Result<()> {
...
-            panic!("TransactionManager failed! {:?}", e);
+            thread_error!("TransactionManager failed! {:?}", e);
+            panic!("Fatal error in TransactionManager!");
         }
     });
     let tc_jito_rx = jito_rx.clone();
     thread::spawn(move || {
         if let Err(e) = transaction_checker.start(tc_jito_rx, pending_bundles) {
-            panic!("TransactionChecker failed! {:?}", e);
+            thread_error!("TransactionChecker failed! {:?}", e);
```
IliaZyrin, line 145: General comment: do you still think it's useful to print the thread id in the logs? I thought it was only for the "getting rid of tokio" part of refactoring.

IliaZyrin (approved):
Reviewed together in a call - LGTM.

LevBeta:
> > Ephemeral caches were part of the original idea for a faster, stateless liquidator that avoids unnecessary overhead. The goal was to reduce duplicated work across a liquidator or cluster of liquidators, where each instance currently processes the same data independently—even when most of that work is redundant. By using short-lived, task-specific caches instead of a full persistent state, we can greatly improve performance, lower memory usage, and make the system more efficient overall.
> 
> Hey hey! Long time no see! 🍻 Thanks for looking into this.
> 
> The cache is in fact ephemeral, not even sure why you think it is persistent. :-) Unlike the original design though it is shareable between processes. Reasoning for it is to first decouple data from processing and to avoid duplicate caches of the same data in different processes. This greatly simplifies maintenance and might slightly improve performance. Another big problem this PR is fixing is the fact that rebalancing and liquidation are running in the same thread that is draining the Geyser messages channels. Since rebalancing and particularly liquidation is time consuming, the channels are growing in size. That is causing liquidations to fail due to outdated oracle and account data and eventually leads to OOM.

So for example, all the mfi accounts are being loaded into the cache right from the start in the cache loader, but I don’t see any method that actually clears them out later. That doesn’t feel very “ephemeral,” to put it simply. From what I can tell, the Clock is being wrapped into the oracles cache, but that’s a totally different use case — that’s more about time compliance logic.

What I’m trying to figure out is: where’s the cache management that keeps this from turning into a bunch of always-running, mostly-useless cached strutcs? Or is that just not handled right now?

src/cli/entrypoints.rs
andreimontchik: Yeah, it was pretty vital when we were chasing down the deadlocking threads. We will most likely clean in up after Eva is stable and healthy.

andreimontchik:
> > > Ephemeral caches were part of the original idea for a faster, stateless liquidator that avoids unnecessary overhead. The goal was to reduce duplicated work across a liquidator or cluster of liquidators, where each instance currently processes the same data independently—even when most of that work is redundant. By using short-lived, task-specific caches instead of a full persistent state, we can greatly improve performance, lower memory usage, and make the system more efficient overall.
> > 
> > 
> > Hey hey! Long time no see! 🍻 Thanks for looking into this.
> > The cache is in fact ephemeral, not even sure why you think it is persistent. :-) Unlike the original design though it is shareable between processes. Reasoning for it is to first decouple data from processing and to avoid duplicate caches of the same data in different processes. This greatly simplifies maintenance and might slightly improve performance. Another big problem this PR is fixing is the fact that rebalancing and liquidation are running in the same thread that is draining the Geyser messages channels. Since rebalancing and particularly liquidation is time consuming, the channels are growing in size. That is causing liquidations to fail due to outdated oracle and account data and eventually leads to OOM.
> 
> So for example, all the mfi accounts are being loaded into the cache right from the start in the cache loader, but I don’t see any method that actually clears them out later. That doesn’t feel very “ephemeral,” to put it simply. From what I can tell, the Clock is being wrapped into the oracles cache, but that’s a totally different use case — that’s more about time compliance logic.
> 
> What I’m trying to figure out is: where’s the cache management that keeps this from turning into a bunch of always-running, mostly-useless cached strutcs? Or is that just not handled right now?

1. Cache is no more after the Eva instance goes down, meaning it is not being persisted and not surviving restarts. There is no need to clean up accounts in cache and honestly I am not quite sure what you mean by this.
2. Different process, not related to the cache is responsible for the Solana clock update. Can you please point to where it being wrapped into the Oracles cache? Could be an overlook from my side. 
3. The GeyserProcessor is responsible for cache update.

PR #31, opened 2025-05-07
Hardening the rebalancing

andreimontchik:
Making the rebalancing process more durable.

### DIFF
--- a/src/cache/mints.rs
+++ b/src/cache/mints.rs
@@ -51,7 +51,7 @@ impl MintsCache {
 
     pub fn try_get_mint_for_token(&self, token_address: &Pubkey) -> Result<Pubkey> {
         self.get_mint_for_token(token_address).ok_or(anyhow!(
-            "Failed ot find Mint for the Token {} in Cache!",
+            "Failed to find Mint for the Token {} in Cache!",
             &token_address
         ))
     }
--- a/src/cache/mints.rs
+++ b/src/cache/mints.rs
@@ -45,14 +45,15 @@ impl MintsCache {
         self.mints.values().map(|mint| mint.token).collect()
     }
 
+    pub fn get_mint_for_token(&self, token_address: &Pubkey) -> Option<Pubkey> {
+        self.token_to_mint.get(token_address).copied()
+    }
+
     pub fn try_get_mint_for_token(&self, token_address: &Pubkey) -> Result<Pubkey> {
-        self.token_to_mint
-            .get(token_address)
-            .ok_or(anyhow!(
-                "Failed ot find Mint for the Token {} in Cache!",
-                &token_address
-            ))
-            .copied()
+        self.get_mint_for_token(token_address).ok_or(anyhow!(
+            "Failed ot find Mint for the Token {} in Cache!",
+            &token_address
+        ))
     }
 
     pub fn len(&self) -> usize {
--- a/src/cache/mints.rs
+++ b/src/cache/mints.rs
@@ -18,42 +18,31 @@ impl MintsCache {
         }
     }
 
-    pub fn try_insert(
-        &mut self,
-        mint_address: Pubkey,
-        mint: Account,
-        token_address: Pubkey,
-    ) -> anyhow::Result<()> {
+    pub fn insert(&mut self, mint_address: Pubkey, mint: Account, token_address: Pubkey) {
         self.mints
             .insert(mint_address, MintWrapper::new(token_address, mint));
         self.token_to_mint.insert(token_address, mint_address);
-        Ok(())
-    }
-
-    pub fn get_account(&self, address: &Pubkey) -> Option<MintWrapper> {
-        self.mints.get(address).cloned()
     }
 
     pub fn try_get_account(&self, address: &Pubkey) -> Result<MintWrapper> {
-        self.get_account(address).ok_or(anyhow!(
-            "Failed ot find Mint for the Address {} in Cache!",
-            &address
-        ))
+        self.mints
+            .get(address)
+            .ok_or(anyhow!("Failed ot find Mint for the Address {}!", &address))
+            .cloned()
     }
 
     pub fn get_tokens(&self) -> Vec<Pubkey> {
         self.mints.values().map(|mint| mint.token).collect()
     }
 
-    pub fn get_mint_for_token(&self, token_address: &Pubkey) -> Option<Pubkey> {
-        self.token_to_mint.get(token_address).copied()
-    }
-
     pub fn try_get_mint_for_token(&self, token_address: &Pubkey) -> Result<Pubkey> {
-        self.get_mint_for_token(token_address).ok_or(anyhow!(
-            "Failed to find Mint for the Token {} in Cache!",
-            &token_address
-        ))
+        self.token_to_mint
+            .get(token_address)
+            .ok_or(anyhow!(
+                "Failed to find Mint for the Token {}!",
+                &token_address
+            ))
+            .copied()
     }
 
     pub fn len(&self) -> usize {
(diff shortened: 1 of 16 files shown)


src/cache/mints.rs
```diff
@@ -45,14 +45,15 @@ impl MintsCache {
...
     pub fn try_get_mint_for_token(&self, token_address: &Pubkey) -> Result<Pubkey> {
-        self.token_to_mint
-            .get(token_address)
-            .ok_or(anyhow!(
-                "Failed ot find Mint for the Token {} in Cache!",
-                &token_address
-            ))
-            .copied()
+        self.get_mint_for_token(token_address).ok_or(anyhow!(
+            "Failed ot find Mint for the Token {} in Cache!",
```
jgur-psyops, line 54: nit: typo

src/cache/oracles.rs
```diff
@@ -8,35 +9,37 @@ use anyhow::{anyhow, Result};
...
-        self.accounts.insert(oracle_address, account);
-        self.account_wrappers
-            .write()
-            .map_err(|e| anyhow!("Failed to lock the account_wrappers map for insert! {}", e))?
-            .insert(oracle_address, account_wrapper);
-        self.oracle_to_bank.insert(oracle_address, bank_address);
-        self.bank_to_oracle.insert(bank_address, oracle_address);
+        self.accounts.insert(address, account);
+        self.oracle_to_bank.insert(address, bank_address);
+        self.bank_to_oracle.insert(bank_address, address);
```
jgur-psyops, line 36: nit: fine for now, but not all banks have one oracle, eg staked collateral banks have 3, kamino will have 2

src/cache/oracles.rs
```diff
@@ -79,66 +70,42 @@ impl<T: OracleWrapperTrait + Clone> OraclesCache<T> {
...
             .get(oracle)
-            .cloned()?;
-        Some(wrapper.clone())
+            .cloned()
     }
     pub fn try_get_wrapper_from_bank(&self, bank_pk: &Pubkey) -> Result<T> {
-        let oracle_pk = self.bank_to_oracle.get(bank_pk).ok_or(anyhow::anyhow!(
-            "Failed ot find the Oracle Pubkey for the Bank {} in Cache!",
+        self.get_wrapper_from_bank(bank_pk).ok_or(anyhow::anyhow!(
+            "Failed ot find the Oracle for Bank {} in Cache!",
```
jgur-psyops, line 92: nit: typo

src/cache/oracles.rs
```diff
@@ -79,66 +70,42 @@ impl<T: OracleWrapperTrait + Clone> OraclesCache<T> {
...
     }
 
     pub fn try_get_bank_from_oracle(&self, oracle_address: &Pubkey) -> Result<Pubkey> {
-        let bank_address = self
-            .oracle_to_bank
+        self.oracle_to_bank
             .get(oracle_address)
+            .cloned()
             .ok_or(anyhow::anyhow!(
                 "Failed ot find the Bank address for the Oracle {} in Cache!",
```
jgur-psyops, line 102: nit: typo

src/cache_loader.rs
```diff
@@ -257,7 +257,7 @@ impl CacheLoader {
                 continue;
             }
 
-            // TODO: what is this for?
+            // TODO: Only one Oracle is used. What about the others?
```
jgur-psyops, line 260: only staked collateral banks currently use more than one.

src/cache_loader.rs
```diff
@@ -345,32 +343,52 @@ impl CacheLoader {
...
-                );
-            }
+            match price_adapter {
+                std::result::Result::Ok(adapter) => {
+                    let oracle = T::new(oracle_address, adapter);
 
-            cache
-                .oracles
-                .try_insert(oracle_account, oracle, *bank_address)?;
+                    cache.oracles.try_insert(
```
jgur-psyops, line 365: I think try insert aborts if the key already exists, I think we would want to over write (iirc entry? is the function). check me if I'm wrong we don't have maps on chain so it's been a while lol

src/rebalancer.rs
```diff
@@ -191,20 +186,31 @@ impl Rebalancer {
...
+            error!("Failed to repay liabilities! {}", error)
+        }
 
-        self.deposit_preferred_tokens()?;
+        if let Err(error) = self.drain_tokens_from_token_accounts() {
+            error!("Failed to drain the Liquidator's tokens! {}", error)
+        }
 
-        Ok(())
+        if let Err(error) = self.deposit_preferred_tokens() {
```
jgur-psyops, line 211: I guess in theory this could make the liquidator go broke if it can't deposit then it has nothing to liquidate against. Will it retry deposit on the next go?

similarly if sell fails it's exposed to delta risk by holding token, does it try again?

src/cache/oracles.rs
andreimontchik: Yep, noticed that, thanks. This would require more digging and will be addresses a bit later.

src/cache/oracles.rs
andreimontchik: Nope, it is legit here: https://doc.rust-lang.org/src/core/option.rs.html#1955

src/cache_loader.rs
```diff
@@ -221,7 +220,7 @@ impl CacheLoader {
         &self,
         cache: &mut CacheT<T>,
     ) -> anyhow::Result<()> {
-        info!("Loading oracles...");
+        info!("Loading Oracles...");
```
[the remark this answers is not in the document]
andreimontchik: Totally agree that the banks to  oracles relationship could have been done simpler, and this is what I am planning to tackle later. I'd be agains hardcoding it though because this would require code change and redeployment for supporting new banks and will not work for Arena.

Created the ticket for the time being: https://linear.app/marginfi/issue/MRGN-115/lq-refactor-the-banks-to-oracles-relationship

src/cache_loader.rs
andreimontchik: Oooo! You have just found a bug I have been chasing last nite! 🙇🏻 Something was poisoning the RWLock and now I know what it was. Thanks!

src/rebalancer.rs
andreimontchik: I do not believe the delta risk has even been considered in the current Liquidator implementation. This is something that is bothering me, but honestly not too much. Rebalancing is usually occurring every few seconds. 

We might be able to address it though, but that might require quite heavy refactoring to make Liquidation/rebalancing operations atomic. If it is possible at all.

src/cache_loader.rs
andreimontchik: Shut, the insert is actually replacing the existing value. https://doc.rust-lang.org/src/core/option.rs.html#1593

src/cache_loader.rs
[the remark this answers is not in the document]
jgur-psyops: Yeah maybe hardcoding is the wrong word but it should fetch the map at startup from some endpoint and we can update just that endpoint when adding a bank to sync it across all of our systems.

You could ask FE to give you oracle at: https://storage.googleapis.com/mrgn-public/mrgn-bank-metadata-cache.json too

src/cache/oracles.rs
andreimontchik: :-). Replaced cloned() with copied() though where available since the latter is a bit faster. Looked into this because of your comment.

src/cache_loader.rs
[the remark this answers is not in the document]
andreimontchik: It currently loads all bank accounts that belong to the Marginfi program and then gets collects Oracle pubkeys from it. Seems to be working fine, but this approach does not support the newly created banks and liquidator should be restarted for this. Do see this as big issue as the moment, but we could look into this after Eva is stable on production.

https://linear.app/marginfi/issue/MRGN-116/lq-consider-load-banks-and-oracles-from-the-marginfi-api-endpoints

IliaZyrin (approved):

PR #32, opened 2025-05-08
Config changes

IliaZyrin:
Removing SetupFromCli and also tx_landing_url, liquidator_account from config.

### DIFF
--- a/src/cli/setup/mod.rs
+++ b/src/cli/setup/mod.rs
@@ -1,11 +1,9 @@
-use super::app::SetupFromCliOpts;
 use crate::{
     config::{Eva01Config, GeneralConfig, LiquidatorCfg, RebalancerCfg},
     utils::{ask_keypair_until_valid, expand_tilde, is_valid_url, prompt_user},
 };
 
 use anyhow::bail;
-use fixed::types::I80F48;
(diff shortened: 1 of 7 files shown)


src/cli/setup/mod.rs
```diff
@@ -62,24 +58,14 @@ pub fn setup() -> anyhow::Result<()> {
...
-        "Select marginfi group [default: {:?}]: ",
-        GeneralConfig::default_marginfi_group_address()
-    ))?;
-    let marginfi_group_address = if input_raw.is_empty() {
-        GeneralConfig::default_marginfi_group_address()
-    } else {
-        Pubkey::from_str(&input_raw).expect("Invalid marginfi group address")
-    };
+    let input_raw = prompt_user(
+        "Select marginfi group [main group: 4qp6Fx6tnZkY5Wropq9wUYgtFxXKwE6viZxFHg3rdAG8]: ",
```
andreimontchik, line 62: You sure you wanna hardcode it? It is different for Staging.

IliaZyrin: It's not hardcoded, it's just a hint for the users to help them set up the config for the first time. I think it makes sense to guide them through the very basic example - prod main pool.

PR #33, opened 2025-05-09
Cache loading improvement

andreimontchik:


src/cache/oracles.rs
```diff
@@ -44,6 +44,38 @@ impl<T: OracleWrapperTrait + Clone> OraclesCache<T> {
...
+                &address
+            ))
+            .cloned()
+    }
+
+    pub fn exists(&self, address: &Pubkey) -> bool {
+        self.try_get_account(address).is_ok()
+    }
+
+    pub fn try_wire_with_bank(
```
IliaZyrin, line 61: Nitpick: I think a common word for such things is "to bind", not "to wire", but if you're used to it, let's keep as is.

IliaZyrin (approved):
Cool stuff!

PR #36, opened 2025-05-12
An improvement for getting txn bundle status

andreimontchik:
Avoid logging unnecessary stack traces for in-progress txns.


IliaZyrin (approved):
Nice!

PR #37, opened 2025-05-12
Liquidation fixes

IliaZyrin:


src/cache_loader.rs
```diff
@@ -332,11 +336,23 @@ impl CacheLoader {
                     let oracle_to_account = |&key| (key, oracle_map.get(&key).unwrap().clone());
 
                     let mut oracle_key_to_account = oracle_to_account(oracle_keys.first().unwrap());
+                    info!(
```
andreimontchik: NIP: wouldn't it be better to combine logging here?

IliaZyrin: It seems I already removed it

andreimontchik (approved):

PR #39, opened 2025-05-14
Disable the Swb price simulation

andreimontchik:


src/liquidator.rs
```diff
@@ -124,7 +124,7 @@ impl Liquidator {
 
     /// Checks if liquidation is needed, for each account one by one
     fn process_all_accounts(&mut self) -> Result<Vec<PreparedLiquidatableAccount>> {
-        self.swb_price_simulator.simulate_swb_prices()?;
+        //        self.swb_price_simulator.simulate_swb_prices()?;
```
IliaZyrin, line 127: If it works this way, shouldn't we just remove this logic completely? Same applies to the commented part below.

andreimontchik: I think simulation might be helpful during the high volatility times. We'd need to overhaul the way Eva is obtaining the most recent prices, I think just subscribing to the right Oracle addresses might be good enough. 
Created the Linear ticket for this: https://linear.app/marginfi/issue/LIQ-16/overhaul-the-way-eva-is-obtaining-recent-oracle-prices

jgur-psyops: subscribing for prices won't work because we don't crank these oracles. 

the price can move and the oracle could be arbitrarily stale, for example it could be a week since the oracle last updated, maybe the price it shows is $20 while the asset is worth $2 today.

If you do want to avoid simulation you could fetch the approximate current price from an off-chain source like birdeye but that's not ideal for liquidation since small differences in price can make the difference on healthy vs not

andreimontchik: What is preventing us from cranking Oracles more often? Also, I do not believe it should be the Liquidator responsibility.

PR #40, opened 2025-05-20
Emode

andreimontchik:


Cargo.toml
```diff
@@ -6,67 +6,68 @@ edition = "2021"
...
-anchor-spl = { git = "https://github.com/mrgnlabs/anchor.git", features = [
-    "token_2022",
-], rev = "fdcf299dc55ecf7cfa8c4d598aecb1363b99c02d" }
-anchor-client = { git = "https://github.com/mrgnlabs/anchor.git", rev = "fdcf299dc55ecf7cfa8c4d598aecb1363b99c02d", features = [
-    "async",
-] }
+solana-account-decoder = "=2.1.20"
+solana-client = "=2.1.20"
+solana-program = "=2.1.20"
+# solana-rpc-client-api = "=2.1.20"
```
IliaZyrin, line 12: If these are not necessary anymore, let's remove them right away

Cargo.toml
```diff
@@ -6,67 +6,68 @@ edition = "2021"
...
-spl-associated-token-account = "2.0.0"
-spl-token = "=4.0.0"
 thiserror = "1.0.56"
 tokio = { version = "1.35.1", features = ["rt", "macros"] }
 toml = "0.8.12"
 tonic = "0.10.2"
 tonic-health = "0.10.2"
-yellowstone-grpc-client = { git = "https://github.com/mrgnlabs/yellowstone-grpc", branch = "1.18.17" }
-yellowstone-grpc-proto = { git = "https://github.com/mrgnlabs/yellowstone-grpc", branch = "1.18.17" }
+yellowstone-grpc-client = { git = "https://github.com/rpcpool/yellowstone-grpc", branch = "v2.1" }
```
IliaZyrin, line 50: At some point we should think about removing all of these fork repos, to not clutter the active projects we're working on.

Cargo.toml
```diff
@@ -6,67 +6,68 @@ edition = "2021"
...
-solana-sdk = "1.18.17"
-solana-address-lookup-table-program = "1.18.17"
-spl-associated-token-account = "2.0.0"
-spl-token = "=4.0.0"
 thiserror = "1.0.56"
 tokio = { version = "1.35.1", features = ["rt", "macros"] }
 toml = "0.8.12"
 tonic = "0.10.2"
 tonic-health = "0.10.2"
-yellowstone-grpc-client = { git = "https://github.com/mrgnlabs/yellowstone-grpc", branch = "1.18.17" }
```
IliaZyrin, line 57: At some point we should think about removing all of these (mrgnlabs) fork repos, to not clutter the active projects we're working on.

src/cache_loader.rs
```diff
@@ -251,15 +232,15 @@ impl CacheLoader {
...
+                thread_error!(
+                    "Discarding Oracles for the non-staked Bank {:?} because it has more than one oracle key, which is not supported.",
                     bank_address
                 );
                 continue;
             }
 
             // Use the first supported Oracle (for non-staked oracles only).
-            let (oracle_address, mut oracle_account) = {
+            match {
```
IliaZyrin, line 243: Nitpick: I would rather keep them as variables without unwrapping and then used `match` on the variables. Otherwise it looks a bit heavy for a `match` statement.

src/cache_loader.rs
```diff
@@ -272,152 +253,205 @@ impl CacheLoader {
...
+                cache.oracles.try_insert(
+                    bank_oracle_address,
+                    bank_oracle.clone(),
+                    None,
+                    *bank_address,
+                )?;
+
+                let lst_mint_oracle_address = *oracle_keys.get(1).unwrap();
+                let mut lst_mint_oracle = oracle_map.get(&lst_mint_oracle_address).ok_or_else(|| {
+                            anyhow!(
```
IliaZyrin, line 394: Nitpick: it seems formatting is broken here.

IliaZyrin (approved):
Nice improvements, especially the deprecated code replacements!

Cargo.toml
IliaZyrin: Nono, I mean physically remove the repos themselves, not only links to them

src/cache_loader.rs
andreimontchik: That is the syntax preference, but I will put it back since it was coded this way originally. 
The purpose of the change is getting rid of unwraps that caused a panic and Eva crash. Let us avoid using them even in the "obvious" places, please.

src/cache_loader.rs
andreimontchik: Using rustfmt.

src/cache_loader.rs
IliaZyrin: Sure sure, I agree with getting rid of unwraps, the comment was only about the huge `match` block.

src/utils/mod.rs
```diff
@@ -469,11 +447,37 @@ pub fn calc_weighted_assets_new(
 }
 
 #[inline(always)]
-pub fn calc_weighted_liabs_new(
+// Copy pasta from https://github.com/mrgnlabs/marginfi-v2/blob/87f1b8fdcde591566ab51e26a3c47554af4bf856/programs/marginfi/src/state/marginfi_account.rs#L322
+// TODO: replace with the on-chain program function call when it becomes available
+fn calculate_bank_asset_weight(
```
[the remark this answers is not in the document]
jgur-psyops: Maybe, it's a bit tricky to edit a section that's just passed audit. If we make `calc_weighted_asset_value` pub will that work? Maybe the same for the other [inline] private functions above it too. 

Anyways feel free to PR into the program to change it for 0.1.4, probably low priority otherwise.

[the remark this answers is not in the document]
andreimontchik: Well, that would require creating the BankAccountWithPriceFeed object, which is a bit tricky. To me it seems easier to wrap the asset weight calculation in a function, I will try to PR it for 0.1.4. Thanks!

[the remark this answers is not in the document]
andreimontchik: Cross-referencing: https://linear.app/marginfi/issue/LIQ-21/wrap-asset-weight-calculation-in-a-public-function-in-the-program

PR #42, opened 2025-05-29
Decoupled Banks from Oracles

andreimontchik:


src/geyser_processor.rs
```diff
@@ -75,72 +66,16 @@ impl<T: OracleWrapperTrait + Clone> GeyserProcessor<T> {
...
-                                )?
-                            }
-                        };
-
-                    self.cache.oracles.try_update_account_wrapper(
-                        &msg.address,
-                        &bank_address,
-                        oracle_price_adapter,
-                    )?;
-                }
```
andreimontchik, line 134: All this no longer needed.

src/utils/swb_cranker.rs
andreimontchik, line 1: I commented simulation functionality for now and will remove when it is confirmed that it is no longer needed.

PR #44, opened 2025-06-04
JITO upgrate related fixes

andreimontchik:


src/transaction_manager.rs
```diff
@@ -40,7 +41,7 @@ pub type BatchTransactions = Vec<RawTransaction>;
 
 pub struct TransactionData {
     pub transactions: BatchTransactions,
-    pub bundle_id: Pubkey,
+    pub account: Pubkey,
```
andreimontchik, line 44: It's not a bundle Id! Cost me few hours of my life to realize it. :-(

src/transaction_manager.rs
```diff
@@ -122,35 +124,46 @@ impl TransactionManager {
...
-                    ERROR_COUNT.inc();
-                    error!("Failed to configure instructions: {:?}", e);
+                    FAILED_LIQUIDATIONS.inc();
+                    thread_error!("Failed to configure instructions: {:?}", e);
                     continue;
                 }
             };
 
-            let bundle = json!(serialized_txs);
+            let bundle = json!([
```
andreimontchik, line 147: An easter egg in the JITO backward incompatible Rust SDK upgrade.

src/transaction_manager.rs
```diff
@@ -122,35 +124,46 @@ impl TransactionManager {
...
                 Err(e) => {
-                    ERROR_COUNT.inc();
-                    error!("Failed to configure instructions: {:?}", e);
+                    FAILED_LIQUIDATIONS.inc();
+                    thread_error!("Failed to configure instructions: {:?}", e);
                     continue;
                 }
             };
 
-            let bundle = json!(serialized_txs);
```
andreimontchik, line 145: JITO conveniently decided to break backward compatibility in their upgraded Rust SDK. :-)

Cargo.toml
```diff
@@ -47,6 +47,7 @@ tonic-health = "0.10.2"
 yellowstone-grpc-client = { git = "https://github.com/rpcpool/yellowstone-grpc", branch = "v2.1" } 
 yellowstone-grpc-proto = { git = "https://github.com/rpcpool/yellowstone-grpc", branch = "v2.1" }
 jito-sdk-rust = "0.2.2"
+base64 = "0.21.0"
```
andreimontchik, line 50: Needed by JITO SDK.

src/transaction_manager.rs
andreimontchik: It is not bundle id and calling it this way confused the hell out of me when I was troubleshooting JITO SDK failures. This is in fact a pubkey of either liquidator or liquidatee account.

IliaZyrin: Ok, let me try once again. If you still disagree with my point, feel free to keep it as-is.
The point is that **it doesn't have to be an account**, it can be anything at all, which uniquely identifies the bundle for the purposes of tracking its current status. Liquidatee account is just one example - since we don't want to repeat the liquidation for the same user when it's already in progress, we use precisely this - that user's ID (liquidatee account). But in rebalancer (and in the other places, potentially in the future) we may want to use something else, rather than just liquidator account.

PR #45, opened 2025-06-06
Removed JITO SDK

andreimontchik:
JITO is proven to be unreliable, so I replaced it with conventional Solana RPC client API. That streamlined the liquidation txn submission and also made the JITO bundle processing functionality obsolete, so I got rid of it. 

It turned out that JITO SDK was hiding the liquidation txn failures and replacing it exposed a great number of previously unreacheable errors. Which is good because now we can tackle them and increase the Eva profitability even further! 😃 

Another improvement is Swb cranking - now it is only happening in case of a  liquidation failure and if the program returned the custom error 0x17a1. :-)
```
> Program log: AnchorError thrown in programs/marginfi/src/state/marginfi_account.rs:408. Error Code: SwitchboardStalePrice. Error Number: 6049. Error Message: Switchboard oracle: stale price.
> Program 
image
[Marginfi V2](https://solscan.io/account/MFv2hWf31Z9kbCa1snEPYctwafyhdvnV7FZnsebVacA) consumed 64627 of 499850 compute units
> Program returned error: custom program error: 0x17a1
```


src/liquidator.rs
```diff
@@ -89,29 +85,55 @@ impl Liquidator {
...
+                                    candidate.liquidatee_account.address,
+                                    e
+                                );
+                                ERROR_COUNT.inc();
+                            }
                         }
-                        let duration = start.elapsed().as_secs_f64();
-                        LIQUIDATION_LATENCY.observe(duration);
                     }
+                    if !swb_oracles.is_empty() {
```
andreimontchik, line 128: Only crank if any of the liquidations failed because of the stale Swb price check.

src/liquidator.rs
```diff
@@ -89,29 +85,55 @@ impl Liquidator {
...
+                        thread_debug!("Cranking Swb Oracles {:#?}", swb_oracles);
```
jgur-psyops, line 129: Particularly with threads, but even if you sent the txes in serial, you can't guarantee the swb crank processes before the liquidation, so some portion of liquidations will fail due to stale oracles. This is fine but it makes it difficult to see when the liquidator is experiencing a genuine oracle error vs transient failure due to the crank not arriving in time.

src/utils/swb_cranker.rs
```diff
@@ -21,25 +43,80 @@ impl Drop for ResetFlag {
...
-            cache,
+            rpc_client,
+            non_blocking_rpc_client,
+            swb_gateway,
+            payer,
             simulation_is_running: Arc::new(AtomicBool::new(false)),
         })
     }
 
+    pub fn crank_oracles(&self, swb_oracles: Vec<Pubkey>) -> Result<()> {
```
jgur-psyops, line 83: Let me see if I understand correctly: this sends a non-blocking rpc to crank the swb oracles, with no preflight, no retries, and without awaiting the signature response? I think that's the best approach if so.

src/liquidator.rs
andreimontchik: Yep, and that is the best approach to balance between excessive cranking and race between liquidations and Geyser price updates I could come up with so far. Since cranking is occurring once after liquidating all candidates, i would put my money on Geyser.

src/wrappers/liquidator_account.rs
```diff
@@ -192,97 +192,80 @@ impl LiquidatorAccount {
...
-                    RpcSendTransactionConfig {
-                        skip_preflight: true,
-                        ..Default::default()
-                    },
-                )?;
-            Ok(())
+                Ok(())
+            }
+            Err(err) => {
+                let mut swb_oracles: Vec<Pubkey> = vec![];
```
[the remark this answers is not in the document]
andreimontchik: Improved. The cranking method caller should log more elaborate txn submission errors now.

src/utils/swb_cranker.rs
andreimontchik: Update to wait for the submission confirmation.

src/liquidator.rs
andreimontchik: Also I added the txn submission confirmation wait. That should give Geyser a leg up.

PR #46, opened 2025-06-09
Using Banks and Oracles LUTs 

andreimontchik:


src/utils/swb_cranker.rs
```diff
@@ -171,11 +171,13 @@ impl SwbCranker {
 }
 
 pub fn is_stale_swb_price_error(err: &ClientError) -> bool {
```
andreimontchik, line 173: Fix for identifying the stale Swb price error.

src/wrappers/liquidator_account.rs
```diff
@@ -215,15 +227,31 @@ impl LiquidatorAccount {
         let recent_blockhash = self
             .rpc_client
             .get_latest_blockhash()
-            .map_err(|e| LiquidationError::new(anyhow!(e)))?;
+            .map_err(|e| LiquidationError::from_anyhow_error(anyhow!(e)))?;
+
+        // Use LUTs only when your transaction involves a large number of observation accounts.
+        let luts: &Vec<AddressLookupTableAccount> = {
+            if total_observation_accounts > 22 {
```
jgur-psyops, line 234: Is there a downside to just always using the LUT? I don't think versioned transactions cost more than legacy so I think you can just always use the LUT version for simplicity.

andreimontchik: They are actually. I had to bump the CU limit from 400_000 to 600_000.  But this is negligible. 
Mac mentioned on the yesterdays AH that "some engineers had issues with using LUTs in the past", that is the main reason for the condition statement. We can reconsider, of course, if this approach causes issues.

PR #48, opened 2025-06-12
Adopting to new infra

andreimontchik:
Refactoring the Eva configuration and adding artifacts for deploying to the new Kubernetes cluster.


src/cli/mod.rs
```diff
@@ -22,55 +18,44 @@ pub mod setup;
 
 /// Main entrypoint for Eva
 pub fn main_entry() -> anyhow::Result<()> {
```
andreimontchik, line 20: No more setup. The configuration comes from env-t vars that are templated in the bin/env.template file.

src/liquidator.rs
```diff
@@ -359,7 +359,11 @@ impl Liquidator {
         let (total_weighted_assets, total_weighted_liabilities) =
             calc_total_weighted_assets_liabs(&self.cache, account, RequirementType::Maintenance)?;
         let maintenance_health = total_weighted_assets - total_weighted_liabilities;
-        thread_debug!("maintenance_health = {:?}", maintenance_health);
+        thread_trace!(
```
andreimontchik, line 362: @IliaZyrin, I added account address to the logging and downgraded it to trace. It is helpful to keep it in DEBUG in production for troubleshooting, but logging this is an overkill.

bin/env.template
```diff
@@ -1,2 +1,27 @@
...
+SIGNER_PUBKEY=<LIQUIDATOR WALLET PUBKEY>
+KEYPAIR_PATH=<LIQIDATOR WALLET FILE>
+COMPUTE_UNIT_PRICE_MICRO_LAMPORTS=1000
+COMPUTE_UNIT_LIMIT=600000
+MARGINFI_PROGRAM_ID=MFv2hWf31Z9kbCa1snEPYctwafyhdvnV7FZnsebVacA
+MARGINFI_GROUPS_WHITELIST=4qp6Fx6tnZkY5Wropq9wUYgtFxXKwE6viZxFHg3rdAG8
+# MARGINFI_GROUPS_BLACKLIST=
+ADDRESS_LOOKUP_TABLES=HGmknUTUmeovMc9ryERNWG6UFZDFDVr9xrum3ZhyL4fC,5FuKF7C1tJji2mXZuJ14U9oDb37is5mmvYLf4KwojoF1,FEFhAFKz48P3w82Ds5VhvyEDwhRqu2FejmnuxEPZ8wNR
+SOLANA_CLOCK_REFRESH_INTERVAL=10
+MIN_PROFIT=0.01
```
jgur-psyops, line 17: This seems low, it barely pays the tx fee and jito tips. I think 10+ cents even up to a dollar is more reasonable.

bin/env.template
```diff
@@ -1,2 +1,27 @@
...
+
+# Liquidator Config
+ISOLATED_BANKS=false
+
+# Rebalancer Config
+TOKEN_ACCOUNT_DUST_THRESHOLD=0.010000000000001563
```
jgur-psyops, line 23: Again this seems somewhat high, a swap on this tiny amount is usually dominated by the tx fee.

bin/env.template
```diff
@@ -1,2 +1,27 @@
...
+LIQUIDATOR_SRC_PATH=<EVA REPOSITORY DIRECTORY>
+
+# General Config
+RPC_URL=https://rpc.ironforge.network/mainnet?apiKey=<IRONFORGE API KEY>
+YELLOWSTONE_ENDPOINT=<ENDPOINT URL>
+YELLOWSTONE_X_TOKEN=<TOKEN>
+SIGNER_PUBKEY=<LIQUIDATOR WALLET PUBKEY>
+KEYPAIR_PATH=<LIQIDATOR WALLET FILE>
+COMPUTE_UNIT_PRICE_MICRO_LAMPORTS=1000
+COMPUTE_UNIT_LIMIT=600000
```
jgur-psyops, line 11: Out of curiosity, I tried our "worst case" 16 position test with this many CU and it was a success, so I can confirm this enough for the worst case!

bin/env.template
IliaZyrin: Agreed. I only set 0.01 when testing specific liquidation errors to 100% catch anything. But it doesn't really make sense to have it this small on prod. $1 seems a better option.

bin/env.template
andreimontchik: This threshold is used to "clean up" the liquidator open positions and token balances. It should only have the USDC asset position open. It is possible, but kinda rare, that the cleanup txn is at loss.

What would be the recommended param value?  I reused the old one.

IliaZyrin (approved):

PR #49, opened 2025-06-13
improvements to not waste SOL

IliaZyrin:


src/wrappers/liquidator_account.rs
```diff
@@ -160,10 +181,19 @@ impl LiquidatorAccount {
             .cache
             .try_get_bank_wrapper(asset_bank)
             .map_err(LiquidationError::from_anyhow_error)?;
+        asset_bank_wrapper
```
andreimontchik, line 184: I think it is better to add this validation to the liquidator.process_account()

src/wrappers/liquidator_account.rs
```diff
@@ -160,10 +181,19 @@ impl LiquidatorAccount {
...
+            .bank
+            .check_utilization_ratio()
+            .map_err(LiquidationError::from_anchor_error)?;
+
         let liab_bank_wrapper = self
             .cache
             .try_get_bank_wrapper(liab_bank)
             .map_err(LiquidationError::from_anyhow_error)?;
+        liab_bank_wrapper
```
andreimontchik, line 193: ditto

src/wrappers/liquidator_account.rs
```diff
@@ -223,6 +262,15 @@ impl LiquidatorAccount {
             liquidatee_account_address,
             liquidatee_observation_accounts
         );
+        if liquidator_swb_oracles
```
andreimontchik, line 265: Shouldn't it be liquidatee_swb_oracles? Also adding the oracle address to logging might help with troubleshooting.

src/wrappers/liquidator_account.rs
```diff
@@ -207,6 +237,15 @@ impl LiquidatorAccount {
             &self.liquidator_address,
             liquidator_observation_accounts
         );
+        if liquidator_swb_oracles
+            .iter()
+            .any(|oracle| stale_swb_oracles.contains(oracle))
+        {
+            thread_warn!(
+                "Stale oracle(s) found among liquidator observation accounts, skipping liquidation attempt..."
```
andreimontchik, line 245: Adding the oracle address to logging might help with troubleshooting.

src/cli/setup/mod.rs
```diff
@@ -171,6 +174,100 @@ pub fn marginfi_account_by_authority(
...
+#[derive(Deserialize, Debug)]
+pub struct Mint {
+    pub address: String,
+    pub decimals: u8,
+    pub name: String,
+    pub symbol: String,
+    pub token_program: String,
+}
+
+const TOTAL_DEPOSITS_USD_THRESHOLD: f64 = 1000.0;
```
IliaZyrin, line 239: @andreimontchik do you think we should make this configurable? Just not sure if it's worth putting it to the general config since it only makes sense for the case when `marginfi_api_key` is not `None`.

PR #52, opened 2025-06-17
Infra update

andreimontchik:


bin/env.template
```diff
@@ -5,8 +5,7 @@ LIQUIDATOR_SRC_PATH=<EVA REPOSITORY DIRECTORY>
 RPC_URL=https://rpc.ironforge.network/mainnet?apiKey=<IRONFORGE API KEY>
 YELLOWSTONE_ENDPOINT=<ENDPOINT URL>
 YELLOWSTONE_X_TOKEN=<TOKEN>
-SIGNER_PUBKEY=<LIQUIDATOR WALLET PUBKEY>
```
andreimontchik, line 8: Signer pubkey is being obtained from the wallet keypair on startup, the way it should have always been. :-)

src/config.rs
```diff
@@ -13,45 +13,32 @@ pub struct Eva01Config {
 impl Eva01Config {
     pub fn new() -> anyhow::Result<Self> {
         //General configuration
-        let rpc_url = std::env::var("RPC_URL")
```
andreimontchik, line 16: Replaced map_err with expect since this is the fatal error that should cause Eva to startup failure.

Cargo.toml
```diff
@@ -61,7 +61,7 @@ tag = "mrgn-0.1.3"
 features = ["mainnet-beta", "client", "no-entrypoint"]
 
 [dev-dependencies]
-tempfile = "=3.20.0"
+serial_test = "3.2.0"
```
andreimontchik, line 64: Needed to make the configuration tests run sequential since they are using the same env-t vars.

src/config.rs
```diff
@@ -191,8 +167,8 @@ impl GeneralConfig {
     pub fn validate(&self) -> anyhow::Result<()> {
         match (!self.marginfi_groups_whitelist.is_empty(), !self.marginfi_groups_blacklist.is_empty()) {
             (true, false) | (false, true) => Ok(()),
-            (true, true) => Err(anyhow::anyhow!("Only one of marginfi_groups_whitelist or marginfi_groups_blacklist must be set.")),
-            (false, false) => Err(anyhow::anyhow!("One of marginfi_groups_whitelist or marginfi_groups_blacklist must be set. marginfi_groups_whitelist must not be empty.")),
+            (true, true) => Err(anyhow::anyhow!("Only one of marginfi_groups_whitelist or marginfi_groups_blacklist must be set")),
+            (false, false) => Err(anyhow::anyhow!("One of marginfi_groups_whitelist or marginfi_groups_blacklist must be set. marginfi_groups_whitelist must not be empty")),
```
IliaZyrin, line 171: So what do we decide on this one? If you want to enforce non-emptiness for both, let's update the hint to reflect that.

andreimontchik: We are still forcing at least one of the lists to be configured. How do you want to update the error messages?

src/wrappers/liquidator_account.rs
```diff
@@ -399,26 +399,31 @@ impl LiquidatorAccount {
...
+        );
 
         let res = self
             .rpc_client
             .send_and_confirm_transaction_with_spinner_and_config(
                 &tx,
                 CommitmentConfig::confirmed(),
                 RpcSendTransactionConfig {
-                    skip_preflight: true,
+                    skip_preflight: false,
```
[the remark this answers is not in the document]
andreimontchik: Yep, it prevents the faulty txn from being sent saving on txn fees.

src/config.rs
IliaZyrin:
```suggestion
            (false, false) => Err(anyhow::anyhow!("One of marginfi_groups_whitelist or marginfi_groups_blacklist must be set and be not empty")),
```

andreimontchik: The PR has already been merged.  I will add it to the next one. Thanks!

PR #53, opened 2025-06-18
Rebalancer fixes

IliaZyrin:


src/rebalancer.rs
```diff
@@ -370,14 +370,21 @@ impl Rebalancer {
         );
 
         // Get the amount of swap token needed to repay the liability
-        let required_swap_token = self.get_amount(liab_usd_value, &self.swap_mint_bank_pk, None)?;
+        let required_swap_token = self.get_amount(
+            liab_usd_value,
+            &self.swap_mint_bank_pk,
+            Some(PriceBias::Low),
```
andreimontchik, line 376: Why was it changed?

.gitignore
```diff
@@ -5,3 +5,4 @@
 config.toml
 .vscode/
 bin/*.env
+*.sh
```
andreimontchik, line 8: NIP: there is the bin directory that is used to store shell scripts.

IliaZyrin: Cool, but the one I added is only relevant for me, so I don't want to commit it anyway. This addition is just a guard for myself to not do that accidentally.

andreimontchik: Right,  right, that is why it is NIP. :-)

PR #54, opened 2025-06-18
Infra fixes

andreimontchik:


src/cli/mod.rs
```diff
@@ -27,14 +32,40 @@ pub fn main_entry() -> anyhow::Result<()> {
...
+        config.general_config.healthcheck_port,
+        health_state.clone(),
+        stop_liquidator.clone(),
+    );
+    thread::spawn(move || healthcheck_server.start());
+
+    *health_state
+        .lock()
+        .expect("Failed to acquire the health_state lock") = HealthState::Running;
+
```
andreimontchik, line 49: Pretty unsophisticated for now, to complete the infra migration. Will most likely be improved in the future, after the Eva multi instance impl is finally no more. :-)

abdulrabbani00 (approved):
LGTM as long as the endpoints you are defining in the deployment are reachable when you run eva locally.

bin/env.template
```diff
@@ -20,6 +20,9 @@ SOLANA_CLOCK_REFRESH_INTERVAL=10
 # Minimum liquidation profit in USDC 
 MIN_PROFIT=1
 
+# Port to check liquidation health state via HTTP. Example: http://localhost:3000/healthz
```
IliaZyrin, line 23:
```suggestion
# Port to check liquidator health state via HTTP. Example: http://localhost:3000/health/
```

Probably typos

src/utils/healthcheck.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub fn start(&self) -> anyhow::Result<()> {
+        thread_info!("Running the Healthcheck server on port {}", self.port);
+
+        let server = Server::http(format!("0.0.0.0:{}", self.port))
+            .map_err(|e| anyhow::anyhow!("Failed to create the Healthcheck server: {}", e))?;
+
+        while !self.stop_liquidator.load(Ordering::Relaxed) {
+            for request in server.incoming_requests() {
+                let response = match request.url() {
+                    "/healthz" => {
```
IliaZyrin, line 69: Oh, so the `z` at the end was intentional. Is it some kind of a convention here?

andreimontchik: Correct.

PR #55, opened 2025-06-18
Harden the token Swap txn

andreimontchik:
Simplified and hardened the token swap txn submission.


src/rebalancer.rs
```diff
@@ -133,12 +131,26 @@ impl Rebalancer {
     pub fn start(&mut self) -> anyhow::Result<()> {
         thread_info!("Starting the Rebalancer loop");
         while !self.stop_liquidator.load(Ordering::Relaxed) {
-            if self.run_rebalance.load(Ordering::Relaxed) && self.needs_to_be_rebalanced()? {
```
andreimontchik, line 136: That was causing the nasty panic.

src/rebalancer.rs
```diff
@@ -604,8 +616,32 @@ impl Rebalancer {
...
+            .rpc_client
+            .send_and_confirm_transaction_with_spinner_and_config(
+                &tx,
+                CommitmentConfig::finalized(),
+                RpcSendTransactionConfig {
+                    skip_preflight: false,
+                    ..Default::default()
+                },
+            )?;
+        thread_info!(
```
IliaZyrin, line 637: This is duplicating most of the info from above. I would just print signature here.

IliaZyrin (approved):
Just a few nitpicks, lgtm!

src/rebalancer.rs
andreimontchik: It was missing oracle for a bank this time. In general, the error should not have been propagated out of the while loop.

src/rebalancer.rs
```diff
@@ -604,8 +616,32 @@ impl Rebalancer {
 
         tx = VersionedTransaction::try_new(tx.message, &[&self.signer])?;
 
-        TransactionSender::aggressive_send_tx(&self.rpc_client, &tx, SenderCfg::DEFAULT)
-            .map_err(|_| anyhow!("Failed to send swap transaction"))?;
+        thread_info!(
```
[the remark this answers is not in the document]
andreimontchik: Let us keep it INFO for now, because Eva is logging INFO in production. This is not gonna happen too often and  we will have mints data in case of failed simulations. We can downgrade logging later if it becomes too noisy.

src/config.rs
```diff
@@ -165,10 +165,17 @@ pub struct GeneralConfig {
...
+        ) {
             (true, false) | (false, true) => Ok(()),
-            (true, true) => Err(anyhow::anyhow!("Only one of marginfi_groups_whitelist or marginfi_groups_blacklist must be set")),
-            (false, false) => Err(anyhow::anyhow!("One of marginfi_groups_whitelist or marginfi_groups_blacklist must be set. marginfi_groups_whitelist must not be empty")),
+            (true, true) => Err(anyhow::anyhow!(
+                "Only one of marginfi_groups_whitelist or marginfi_groups_blacklist must be set."
+            )),
+            (false, false) => Err(anyhow::anyhow!(
+                "Either marginfi_groups_whitelist or marginfi_groups_blacklist must be set - not both."
+            )),
```
andreimontchik, line 178: Improved comments

PR #58, opened 2025-06-27
Sequentically run rebalancing after successful liquidations

andreimontchik:
1. Made rebalancing sequential to the liquidation process. That simplified the whole liquidation -> rebalancing process.
2. Added shutdown hooks for graceful shutdown.
3. Improved the Liquidator account creation for new Arena groups.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -105,8 +105,6 @@ impl Eva01Config {
             .parse()
             .expect("Invalid TOKEN_ACCOUNT_DUST_THRESHOLD number");
 
-        let preferred_mints: Vec<Pubkey> = parse_pubkey_list("PREFERRED_MINTS")?;
-
         let swap_mint = Pubkey::from_str(
             &std::env::var("SWAP_MINT").expect("SWAP_MINT environment variable is not set"),
         )
@@ -122,7 +120,6 @@ impl Eva01Config {
 
         let rebalancer_config = RebalancerCfg {
             token_account_dust_threshold: I80F48::from_num(token_account_dust_threshold),
-            preferred_mints,
             swap_mint,
             jup_swap_api_url,
             slippage_bps,
@@ -253,7 +250,6 @@ impl std::fmt::Display for LiquidatorCfg {
 #[derive(Debug, Default, Clone)]
 pub struct RebalancerCfg {
     pub token_account_dust_threshold: I80F48,
-    pub preferred_mints: Vec<Pubkey>,
     pub swap_mint: Pubkey,
     pub jup_swap_api_url: String,
     pub slippage_bps: u16,
@@ -266,12 +262,10 @@ impl std::fmt::Display for RebalancerCfg {
             "Rebalancer Config: \n\
                 - Token account dust threshold: {}\n\
                 - Swap mint: {}\n\
-                - Preferred mints: {:?}\n\
                 - Jup Swap Api URL: {}\n\
                 - Slippage bps: {}\n",
             self.token_account_dust_threshold,
             self.swap_mint,
-            self.preferred_mints,
             self.jup_swap_api_url,
             self.slippage_bps,
         )
--- a/src/geyser.rs
+++ b/src/geyser.rs
@@ -97,6 +97,7 @@ impl GeyserService {
         thread_info!("Staring GeyserService.");
 
         let tracked_accounts_vec: Vec<Pubkey> = self.tracked_accounts.keys().copied().collect();
+        let tls_config = ClientTlsConfig::new().with_native_roots();
 
         while !self.stop.load(Ordering::Relaxed) {
             thread_info!("Connecting to Geyser...");
@@ -107,7 +108,7 @@ impl GeyserService {
             let mut client = self.tokio_rt.block_on(
                 GeyserGrpcClient::build_from_shared(self.endpoint.clone())?
                     .x_token(self.x_token.clone())?
-                    .tls_config(ClientTlsConfig::new().with_native_roots())?
+                    .tls_config(tls_config.clone())?
                     .connect(),
             )?;
 
--- a/src/liquidator.rs
+++ b/src/liquidator.rs
@@ -1,7 +1,8 @@
 use crate::{
     cache::Cache,
-    config::{GeneralConfig, LiquidatorCfg},
-    metrics::{ERROR_COUNT, FAILED_LIQUIDATIONS, LIQUIDATION_LATENCY},
+    config::Eva01Config,
+    metrics::{ERROR_COUNT, FAILED_LIQUIDATIONS, LIQUIDATION_ATTEMPTS, LIQUIDATION_LATENCY},
+    rebalancer::Rebalancer,
     thread_debug, thread_error, thread_info, thread_trace,
     utils::{calc_total_weighted_assets_liabs, get_free_collateral, swb_cranker::SwbCranker},
     wrappers::{
@@ -32,8 +33,9 @@ use std::{sync::atomic::Ordering, thread};
 
 #[allow(dead_code)]
 pub struct Liquidator {
-    liquidator_account: LiquidatorAccount,
-    config: LiquidatorCfg,
+    liquidator_account: Arc<LiquidatorAccount>,
+    rebalancer: Rebalancer,
+    support_isolated_banks: bool,
     min_profit: f64,
     run_liquidation: Arc<AtomicBool>,
     stop_liquidator: Arc<AtomicBool>,
@@ -51,20 +53,23 @@ pub struct PreparedLiquidatableAccount {
 
 impl Liquidator {
     pub fn new(
-        general_config: GeneralConfig,
-        liquidator_config: LiquidatorCfg,
-        liquidator_account: LiquidatorAccount,
+        config: Eva01Config,
+        liquidator_account: Arc<LiquidatorAccount>,
         run_liquidation: Arc<AtomicBool>,
         stop_liquidator: Arc<AtomicBool>,
         cache: Arc<Cache>,
     ) -> Result<Self> {
-        let swb_cranker = SwbCranker::new(&general_config)?;
+        let swb_cranker = SwbCranker::new(&config.general_config)?;
+
+        let rebalancer =
+            Rebalancer::new(config.clone(), liquidator_account.clone(), cache.clone())?;
 
         Ok(Liquidator {
-            config: liquidator_config,
-            min_profit: general_config.min_profit,
-            run_liquidation,
             liquidator_account,
+            rebalancer,
+            support_isolated_banks: config.liquidator_config.isolated_banks,
+            min_profit: config.general_config.min_profit,
+            run_liquidation,
             stop_liquidator,
             cache,
             swb_cranker,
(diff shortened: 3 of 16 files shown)


src/wrappers/liquidator_account.rs
```diff
@@ -91,37 +93,42 @@ impl LiquidatorAccount {
...
-
-            *new_liquidator_account = Some(liquidator_marginfi_account);
+            while cache
+                .marginfi_accounts
+                .try_get_account(&liquidator_marginfi_account)
+                .is_err()
+            {
+                thread_info!("Waiting for the new account info to arrive...");
+                thread::sleep(Duration::from_secs(5));
+            }
```
andreimontchik, line 117: When new account is created it will be received via Geyser. This ensures that the new account information was received.

src/liquidator.rs
```diff
@@ -130,7 +146,14 @@ impl Liquidator {
                     };
                 }
 
-                thread_debug!("The Liquidation process is complete.");
+                thread_info!("The Liquidation process is complete.");
+
+                if (LIQUIDATION_ATTEMPTS.get() - lq_attempts) > 0f64 {
```
andreimontchik, line 151: Only run rebalancing when there were successful liquidations.

src/main.rs
```diff
@@ -58,8 +66,20 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+    let stop = Arc::new(AtomicBool::new(false));
+    signal_hook::flag::register(SIGINT, stop.clone()).unwrap();
+    signal_hook::flag::register(SIGTERM, stop.clone()).unwrap();
+
+    let stop_hook = Arc::clone(&stop);
+    ctrlc::set_handler(move || {
+        stop_hook.store(true, Ordering::SeqCst);
+        println!("Received stop signal");
+    })
+    .expect("Error setting Ctrl-C handler");
```
andreimontchik, line 79: Registered the hooks for graceful shutdown.

src/marginfi_ixs.rs
```diff
@@ -237,6 +237,7 @@ pub fn initialize_marginfi_account(
     marginfi_group: Pubkey,
     signer_keypair: &Keypair,
 ) -> anyhow::Result<Pubkey> {
+    //TODO: confirm that it should not be PDA
     let marginfi_account_key = Keypair::new();
```
andreimontchik, line 241: This! ☝🏻

src/liquidator.rs
jgur-psyops: wouldn't this skip if the last attempt is a success? ie if attempts = lq attempts

src/marginfi_ixs.rs
jgur-psyops: it's not a PDA, accounts are indeed a keypair.

src/rebalancer.rs
```diff
@@ -106,18 +110,18 @@ impl Rebalancer {
     pub fn fund_liquidator_account(&mut self) -> anyhow::Result<()> {
         // Multiply by 40 because the profit is 0.025 of the amount that is being liquidated.
         // And then multiply by MIN_LIQUIDATIONS_COUNT to get enough funds for that many liquidations.
-        let usd_value = (self.general_config.min_profit * 40.0) * (MIN_LIQUIDATIONS_COUNT as f64);
+        let usd_value = (self.min_profit * 40.0) * (MIN_LIQUIDATIONS_COUNT as f64);
```
jgur-psyops, line 113: what's with magic 40 here let's make it a constant

src/rebalancer.rs
```diff
@@ -106,18 +110,18 @@ impl Rebalancer {
...
 
         let sol_price = 150.0; // TODO: Fetch the current SOL price and account for slippage by lowering it
```
jgur-psyops, line 115: why don't we just fund a static amount in sol why do we care about the sol price

src/rebalancer.rs
```diff
@@ -496,9 +457,8 @@ impl Rebalancer {
         Ok(false)
     }
 
-    fn has_non_preferred_deposits(&self, lq_account: &MarginfiAccountWrapper) -> Result<bool> {
+    fn has_non_preferred_deposits(&self, lq_account: &LendingAccount) -> Result<bool> {
         Ok(lq_account
-            .lending_account
             .balances
             .iter()
             .filter(|balance| balance.is_active())
```
jgur-psyops, line 464: I understand the goal here is ensure the liquidator never has non preferred positions, but a full balance here if the liquidator just grabbed ~20 cents is excessive (I guess it withdraw and abort jup swap due to dust threshold?)

if the goal is to simplify liquidator operation so it never has a balance that isn't preferred a better solution would be to compose a withdraw at the end of liquidation, directly into the tx, and keep balancing those token accounts at regular intervals with a higher dust threshold

jgur-psyops (commented):
mechanically lgtm, but this is a bad idea overall. balancing is expensive (tx cost, thread time) and you can achieve the effect of getting rid of non preferred assets by simply withdrawing after the liquidate instruction.

src/rebalancer.rs
```diff
@@ -51,53 +43,65 @@ const MIN_LIQUIDATIONS_COUNT: u64 = 10;
 /// "rebalanced" -> Document this better
 #[allow(dead_code)]
```
IliaZyrin, line 44: Same here - shouldn't this annotation be removed?

src/rebalancer.rs
IliaZyrin: The comment right above explains this magic. But we can factor it out as well.

src/rebalancer.rs
IliaZyrin: Yeah, so the idea was to always fund enough for `MIN_LIQUIDATIONS_COUNT`. I don't mind to simplify it though, by saying e.g. always transfer 1 SOL equivalent.
The initial goal was to make it less random and bind to some specific expectations (i.e. count of the minimal liquidations we guarantee to be able to perform) but probably not worth it. @andreimontchik what do you think?

src/liquidator.rs
IliaZyrin: Plus it may be beneficial to move `LIQUIDATION_ATTEMPTS` increment further down in the `liquidate()` function so that it catches all the errors up to the point of sending the tx.

IliaZyrin:
> mechanically lgtm, but this is a bad idea overall. balancing is expensive (tx cost, thread time) and you can achieve the effect of getting rid of non preferred assets by simply withdrawing after the liquidate instruction.

This PR doesn't change this aspect, it only streamlines what was already working this "expensive" way. I agree though that we need to improve the rebalancing and make it more efficient in terms of cost and thread time.

src/rebalancer.rs
andreimontchik: I believe this is temporary and will be replaced by more flexible account funding for liquidation purposes.

src/rebalancer.rs
andreimontchik: 1 SOL might not always be enough. I think the better idea might be to fund the Arena group specific liquidator account from the Marginfi groups USDC asset position, liquidate and rebalance it back after the liquidation is completed.

src/rebalancer.rs
andreimontchik: Well, the PR is making rebalancing dependent on liquidation attempts, so it only runs when there was or might have been token balance change. It was excessive and more complicated to run it independently of liquidations.

We can definitely revisit and improve the rebalancing logic going forward though.

src/liquidator.rs
IliaZyrin: Well, `needs_to_be_rebalanced()` will anyway return `false` in this case and rebalancing won't happen, so not a big deal. `check_liquidator_health()` will happen though, but that may be beneficial.

So we can keep it as is, just wanted to point out that there may be dummy rebalancing runs which do nothing.

IliaZyrin (approved):

src/config.rs
```diff
@@ -105,8 +105,6 @@ impl Eva01Config {
             .parse()
             .expect("Invalid TOKEN_ACCOUNT_DUST_THRESHOLD number");
 
-        let preferred_mints: Vec<Pubkey> = parse_pubkey_list("PREFERRED_MINTS")?;
```
andreimontchik, line 108: A cleanup, this param was no longer in use.

src/geyser.rs
```diff
@@ -97,6 +97,7 @@ impl GeyserService {
         thread_info!("Staring GeyserService.");
 
         let tracked_accounts_vec: Vec<Pubkey> = self.tracked_accounts.keys().copied().collect();
+        let tls_config = ClientTlsConfig::new().with_native_roots();
```
andreimontchik, line 100: Geyser was failing to reconnect because of the obscure lazy init.

src/rebalancer.rs
```diff
@@ -128,45 +131,37 @@ impl Rebalancer {
...
-            .marginfi_accounts
-            .try_get_account(&self.liquidator_account.liquidator_address)?;
-
-        self.should_stop_liquidator(&lq_account)?;
-
-        Ok(!self.stop_liquidator.load(Ordering::Relaxed)
-            && self.run_rebalance.load(Ordering::Relaxed)
-            && (self.has_non_preferred_deposits(&lq_account)? || lq_account.has_liabs()))
+    fn needs_to_be_rebalanced(&mut self, lq_acc: &MarginfiAccountWrapper) -> Result<bool> {
+        Ok(self.has_non_preferred_tokens()?
```
andreimontchik, line 162: Fix for the fresh 🐛 - Eva was not evaluating the non preferred tokens balances for deciding whether or not it is a time to rebalance.

src/rebalancer.rs
```diff
@@ -217,89 +234,68 @@ impl Rebalancer {
...
                 Err(error) => {
                     thread_error!(
-                        "Failed to withdraw and sell deposit for the Bank ({}): {:?}",
+                        "Failed to withdraw and sell asset for the Bank ({}): {:?}",
                         bank_pk,
                         error
                     );
                 }
                 Ok(mint_pk) => {
-                    sold_deposits.push((mint_pk, bank_pk));
```
andreimontchik, line 297: Not  so fresh 🐛  - the method was returning the (mint, bank) PK pairs, but the (mint, token) pairs were expected.

src/rebalancer.rs
```diff
@@ -456,7 +464,7 @@ impl Rebalancer {
         Ok(false)
     }
 
-    fn has_non_preferred_deposits(&self, lq_account: &LendingAccount) -> Result<bool> {
+    fn has_non_preferred_assets(&self, lq_account: &LendingAccount) -> Result<bool> {
```
IliaZyrin, line 467: I liked the previous name more, cause it's more relevant to the banks. Assets are too abstract IMHO.

andreimontchik: Sure thing, rolled back. I am used to using assets in this case, that is why I renamed it.

src/rebalancer.rs
```diff
@@ -365,41 +361,43 @@ impl Rebalancer {
...
-            &self.swap_mint_bank_pk,
-            Some(PriceBias::Low),
-        )?;
+        let required_swap_token =
+            self.get_amount(liab_usd_value, &self.swap_mint_bank, Some(PriceBias::Low))?;
 
         let withdraw_amount = if required_swap_token.is_positive() {
-            let (max_withdraw_amount, withdraw_all) =
-                self.get_max_withdraw_for_bank(&self.swap_mint_bank_pk, lq_account)?;
+            // The withdral_all flag is purposefully ignored here.
```
andreimontchik, line 388: The fix to prevent Eva from occasionally draining the preferred asset.

IliaZyrin: I think we should still dig into why it is even trying to withdraw a preferred deposit. But in any case this is for another PR. This one has been growing for too long already, let's merge it in!

andreimontchik: That was an attempt to use the preferred token to pay the newly acquired liability, which is the way it supposed to be, I believe.

IliaZyrin: Ah, true, then we probably should move this step after the other ones, so that we always have enough collateral. But again, for the next PR, this looks good!

IliaZyrin: Just one little typo:
```suggestion
            // The withdraw_all flag is purposefully ignored here.
```

andreimontchik: > Ah, true, then we probably should move this step after the other ones, so that we always have enough collateral. But again, for the next PR, this looks good!

I envision that we will eventually revising the way the liquidation/rebalancing transactions are being created. Some of them could be combined.

PR #59, opened 2025-07-08
Print the Banks and Oracles configurations

andreimontchik:
The change is purely for troubleshooting purposes and not very helpful in production, that is why I wrapped it in the compilation feature flags, which I usually despise. :-)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,3 +69,7 @@ serial_test = "3.2.0"
 [profile.release]
 opt-level = 3
 lto = true
+
+[features]
+print_oracles = []
+print_banks = []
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -69,3 +69,7 @@ serial_test = "3.2.0"
 [profile.release]
 opt-level = 3
 lto = true
+
+[features]
+print_oracles = []
```
jgur-psyops, line 74: seems like just one "print_logs" would suffice but I guess if you want two features then up to you!

jgur-psyops (approved):
lgtm

Cargo.toml
andreimontchik: Yeah, I thought about a shared feature flag, but decided to have two for now. In case if we only need to print either banks or oracles. No strong opinion though, but kinda being lazy to change it. :-) Maybe later, if we end up printing both all the time.

PR #61, opened 2025-07-14
0.1.4 support

IliaZyrin:


src/utils/mod.rs
```diff
@@ -408,40 +406,35 @@ pub fn calc_weighted_bank_liabs(
...
+        {
             let feed_id = bank_config.get_pyth_push_oracle_feed_id().unwrap();
-            vec![
-                PythPushOraclePriceFeed::find_oracle_address(
-                    PYTH_PUSH_PYTH_SPONSORED_SHARD_ID,
-                    feed_id,
-                )
-                .0,
-            ]
+            vec![PythPushOraclePriceFeed::find_oracle_address(PYTH_SPONSORED_SHARD_ID, feed_id).0]
```
andreimontchik, line 430: Please confirm that the bank config will contain the Pyth feed Id, not the on-chain feed price address. I was under impression that it is gonna be the latter.

jgur-psyops: you are correct, it'll be the latter. so if migrated, no funny business finding the oracle from the feed ID, just read it directly from oracles[0]

src/rebalancer.rs
```diff
@@ -708,6 +708,7 @@ impl Rebalancer {
         let price = oracle_wrapper.get_price_of_type(
             marginfi::state::price::OraclePriceType::RealTime,
             price_bias,
+            bank_wrapper.bank.config.oracle_max_confidence,
```
jgur-psyops, line 711: general note: not sure if we care about confidence when balancing or in evaluation

src/utils/mod.rs
andreimontchik: Yeah, I overlooked that the condition statement is part of the the match pattern. 😳 Thanks! 

@IliaZyrin, would it be possible to create unit tests for the `find_oracle_keys()` method?

andreimontchik (commented):
Second attempt to remove my own request to change.

src/utils/mod.rs
IliaZyrin: Yep, will do!

src/rebalancer.rs
IliaZyrin: Good point, replaced with `u32::MAX`.

PR #62, opened 2025-07-23
Simulate and crank stale Swb oracles

andreimontchik:
TLDR: sim stale Swb prices and use them for liquidation evaluation. Crank oracles instead of sending liquidation txn for account that are liquidable with simulated prices. Liquidation of these accounts will be attempted shortly after the cranking is completed successfully, assuming the accounts are still liquidable with actual prices.

### DIFF
--- a/src/utils/swb_cranker.rs
+++ b/src/utils/swb_cranker.rs
@@ -128,21 +128,6 @@ impl SwbCranker {
         Ok(())
     }
 
-    fn calculate_sim_price(mut numbers: Vec<f64>) -> Option<f64> {
-        if numbers.is_empty() {
-            return None;
-        }
-
-        numbers.sort_by(|a, b| a.partial_cmp(b).unwrap());
-        let mid = numbers.len() / 2;
-
-        if numbers.len() % 2 == 0 {
-            Some((numbers[mid - 1] + numbers[mid]) / 2.0)
-        } else {
-            Some(numbers[mid])
-        }
-    }
-
     fn find_stale_oracles(&self, cache: &Cache) -> Result<Vec<Pubkey>> {
         let clock = &clock_manager::get_clock(&cache.clock)?;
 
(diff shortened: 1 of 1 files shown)


bin/env.template
```diff
@@ -17,6 +18,10 @@ MARGINFI_GROUPS_WHITELIST=4qp6Fx6tnZkY5Wropq9wUYgtFxXKwE6viZxFHg3rdAG8
 ADDRESS_LOOKUP_TABLES=HGmknUTUmeovMc9ryERNWG6UFZDFDVr9xrum3ZhyL4fC,5FuKF7C1tJji2mXZuJ14U9oDb37is5mmvYLf4KwojoF1,FEFhAFKz48P3w82Ds5VhvyEDwhRqu2FejmnuxEPZ8wNR
 SOLANA_CLOCK_REFRESH_INTERVAL=10
 
+SWB_PROGRAM_ID=A43DyUGA7s8eXPxqEjJY6EBu1KKbNgfxF8h17VAHn13w
+SWB_CROSSBAR_URL=https://crossbar.switchboard.xyz
+SWB_CROSSBAR_VERBOSE=false
+
```
andreimontchik, line 24: This should have been parametrized a long time ago. Will help us to switch to our own Swb crossbar as soon as it is available.

jgur-psyops: FYI I'm sure you've already seen they are rate-limiting the public crossbar so we are going to have to deploy that soon, likely before this update to eva drops.

src/liquidator.rs
```diff
@@ -93,6 +93,10 @@ impl Liquidator {
                 thread_info!("Running the Liquidation process...");
                 self.run_liquidation.store(false, Ordering::Relaxed);
 
+                if let Err(err) = self.swb_cranker.simulate_stale_prices(&self.cache) {
```
jgur-psyops, line 96: I don't think this approach where you sim just the stale prices is ideal. I would sim all swb prices on each tick and crank all switchboard oracles involved with the account(s) on each liquidation attempt. This seems easier logically than a fallback system, you would no longer have to read swb prices on chain at all.

src/utils/swb_cranker.rs
```diff
@@ -60,24 +72,115 @@ impl SwbCranker {
...
+            crossbar_client,
             swb_gateway,
             payer,
         })
     }
 
+    pub fn simulate_stale_prices(&self, cache: &Cache) -> Result<()> {
+        thread_info!("Simulating stale Swb prices...");
+
+        let stale_oracles = self.find_stale_oracles(cache)?;
```
jgur-psyops, line 101: Again re: above comment, it's easier logically to just sim all the swb oracles. Simming just the stale ones opens this up to race conditions e.g. another goes stale as this is running, etc.

src/utils/swb_cranker.rs
```diff
@@ -60,24 +72,115 @@ impl SwbCranker {
...
+            }
+        }
+
+        cache.oracles.try_update_simulated_prices(sim_prices)?;
+        thread_info!("Simulation for the stale Swb prices completed.");
+
+        Ok(())
+    }
+
+    fn calculate_sim_price(mut numbers: Vec<f64>) -> Option<f64> {
```
jgur-psyops, line 131: I don't understand why this exists, is it being called anywhere outside of tests?

src/liquidator.rs
jgur-psyops: I think it was error prone because you were not cranking all oracles and trying to account for staleness while selectively cranking just the stale ones. There isn't really much that can error if you always crank all of the swb oracles prior to a liquidation event: either the crank lands before the tx or it doesn't. 

Every evaluation tick should use the last-simulated prices for swb or on-chain prices for pyth.

Every liquidation event should follow the process of : crank all swb oracles on the account -> read the updated prices on-chain -> liquidate.

andreimontchik: The cranking is time consuming, faulty at times and might become costly if we excessively do for all swb oracles of liquidable accounts. 
We can get back to it though if the selective cranking of the stale oracles will not work as expected.

src/utils/swb_cranker.rs
andreimontchik: Leftovers, gonna remove them, thanks.

PR #64, opened 2025-08-06
Triton RPC related fixes

andreimontchik:
1. Replaced the Ironfogre URL with Triton in the env template file.
2. Removed the leaked RPC keys from logging.
3. Fixed the faulty Swb cranking.
4. Upgraded the protocol version to 0.1.4-rc3


src/config.rs
```diff
@@ -196,9 +196,7 @@ impl std::fmt::Display for GeneralConfig {
         write!(
             f,
             "GeneralConfig:\n\
-                 - RPC URL: {}\n\
```
IliaZyrin, line 199: Is this to not expose the secrets?

PR #67, opened 2025-09-16
fix for rebalancing and health publishings

IliaZyrin:


abdulrabbani00 (commented):
Works but I am worried about individual writes.

IliaZyrin:
> Also, is it possible to run eva without writing to the DB? I was thinking we could run two instance, one for writing and one for liquidating.
> 
> Your call here. But just a concern, some integrators might not want to plug this into a DB or might want to turn off the DB Write.

- Yep I introduced [a new feature](https://github.com/mrgnlabs/eva01/pull/67/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R7) for this - `publish_to_db` - which can be turned on or off as you wish. It regulates both writing to DB and (intentionally not removed for now) writing to a CSV file. We can totally run two instances only one of which would be publishing so that another one is fully "focused" on liquidations.

PR #76, opened 2025-12-08
Basic Swapping with titan

abdulrabbani00:


src/titan/mod.rs
```diff
@@ -0,0 +1,206 @@
...
+
+/// Configuration for Titan swap client
+pub struct TitanSwapConfig {
+    /// Solana RPC endpoint URL
+    pub rpc_url: String,
+    /// Titan WebSocket endpoint (e.g., "us1.api.demo.titan.exchange")
+    pub titan_ws_endpoint: String,
+    /// Titan API key (required for direct mode)
+    pub titan_api_key: Option<String>,
+    /// Hermes proxy endpoint (alternative to direct mode)
```
IliaZyrin: Why do we need it? Aren't we always gonna set the titan API key?

abdulrabbani00: Don't really need it, for now I am setting to none. But maybe something we will use down the road so might as well leave it in.

abdulrabbani00:
Made a few minor changes if you want to review. Essentially plugging the Titan swap into the rebalancer in a very rudimentary way:

1. Pick Titan always. If it fails, fall back to Jupiter.
2. For Titan swap, create and close the client for every swap.

I just wanted to make a little progress on this. Some open questions I have that will inform the next iteration:

- How often are we swapping, and is swap speed incredibly sensitive? It would be great to sim the two aggregators and see who can give us a better rate, then go with that. Obviously this requires a little more time, but it could be worthwhile.
- If we do a ton of swaps, then perhaps creating and closing a WS per swap is overkill. Thoughts?


@IliaZyrin

PR #93, opened 2026-05-27
feat: liquidator improvements - oracle handling, rebalancer stability, integration account refresh

k0beLeenders:
### Oracle & integration simulation

Removed per-scan-cycle simulate bundle for oracle & integration simulation.

Replaced swb oracle simulation with periodic price fetch using `/v0/realprice` every 30s (API → Crossbar fallback) and writes synthetic `PullFeedAccountData` accounts to `cache.oracles` so the SDK staleness check always passes.

Replaced integration simulation with periodic integration refresh (`IntegrationAccountFetcher`).

Fetches real accounts via RPC every 30 minutes and patches staleness timestamp/slot bytes in-place, removing all simulation.

---

### Reactive-only crank behavior

Crank transactions were firing unconditionally before every liquidation.

The real cause is a bug where the account is unhealthy locally but healthy onchain causing the liquidation to fail.

This bug will be tracked down and fixed in another PR.

In the meantime, to avoid the oracles from getting crank-looped (because the previous sequence is getting triggered very often), I changed this to reactive-only crank behavior.

This means we only crank again when the onchain liquidation fails with `StaleOracles`.

---

### Rebalance stale oracle handling

Rebalances were blocked by stale oracle errors.

We can be more lenient here since oracle freshness is not a requirement for rebalancing and we only need approximate prices for swap routing.

---

### Declared value range

Removed declared value range.

If the manual range was not updated in time this could block legitimate liquidations.

---

### LUT fallback & refresh

If the LUT is invalid or deactivated the flow was fully blocked.

It now detects this automatically and retries without LUTs.

Also added a subcommand to create/refresh LUTs.  

LUTs are now split into 3 asset-based groups with smarter selection based on the full liquidation position, plus a lightweight overflow retry path for oversized transactions. The refresh flow is also resume-safe and can extend partially-created LUTs automatically.

---

### Missing token account handling

During liquidation health checks the flow could panic if a token account was missing.

This is now handled correctly and includes full logging for easier debugging.

---

New env var: 
PROJECT_0_API_URL

ADDRESS_LOOKUP_TABLES_GROUP1=AXkqQ68ZeAubcs9GLWucuMNHBJcFLFS1C5fx6TtqCubh,2tBWxsFMGrxti8ahu2iNLRXDdwwQrZzjjDKPFtzTCy7L,84zfB8vdzofoQaNkNfDcALsviQpj6VzFgcSettDHRf4q,7AbN7DH7Rrqbga2jUVERji6yN8xPU4zPYRmQGzX4u3ek                   
                                                 
ADDRESS_LOOKUP_TABLES_GROUP2=9QZCeqMAiCPQgMcvi5FATrAVrbNxxDSKZTp2RuJNJsAR,3ZRbZDPnkehkkvZ1ATm3yA6Zqdu2ZDeusDDF55RyWppD 
                                                                                                                                                             
ADDRESS_LOOKUP_TABLES_GROUP3=GF3eTWgCzTH9UPaATvmB6GJnnxuPVeRno316Dzk6pLFd,8zTeUZGes8q2yL18nMPSTDGPDLNfknD2ngehiiy6j51Z


src/cache/banks.rs
```diff
@@ -61,6 +61,22 @@ impl BanksCache {
             .collect()
     }
 
+    /// Returns a map of SwitchboardPull oracle pubkey → bank address.
+    /// Only includes direct SwitchboardPull banks (not integration types), since
+    /// those are the only ones whose prices are stored and read via swb_prices cache.
+    pub fn get_swb_oracle_to_bank_map(&self) -> HashMap<Pubkey, Pubkey> {
```
IliaZyrin: One oracle can be shared among several banks (e.g. P0 USDC, Kamino USDC, etc.)

src/utils/swb_price_fetcher.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+#[derive(Deserialize)]
+struct ViewPriceEntry {
+    #[serde(rename = "oraclePrice")]
+    oracle_price: OraclePriceDto,
+}
+
+#[derive(Deserialize)]
+struct OraclePriceDto {
```
IliaZyrin, line 69: Nitpick but I would use:
```
#[serde(rename_all = "camelCase")]
struct OraclePriceDto {
```

src/utils/swb_price_fetcher.rs
```diff
@@ -0,0 +1,222 @@
...
+    price_realtime: PriceWithConfidenceDto,
+    #[serde(rename = "priceWeighted")]
+    price_weighted: PriceWithConfidenceDto,
+}
+
+#[derive(Deserialize)]
+struct PriceWithConfidenceDto {
+    price: String,
+    confidence: String,
+}
```
IliaZyrin, line 87: I suggest moving parsing logic directly to the struct:

```
#[derive(Debug, Deserialize)]
#[serde(rename_all = "camelCase")]
struct PriceWithConfidenceDto {
    #[serde(deserialize_with = "deserialize_f64_from_string")]
    price: f64,

    #[serde(deserialize_with = "deserialize_f64_from_string")]
    confidence: f64,
}

// Can be placed in utils and reused
fn deserialize_f64_from_string<'de, D>(deserializer: D) -> Result<f64, D::Error>
where
    D: Deserializer<'de>,
{
    let s = String::deserialize(deserializer)?;
    s.parse::<f64>().map_err(serde::de::Error::custom)
}
```

src/utils/swb_price_fetcher.rs
```diff
@@ -0,0 +1,222 @@
...
+        crossbar_api_url: Option<String>,
+        cache: Arc<Cache>,
+        stop: Arc<AtomicBool>,
+    ) -> Self {
+        let crossbar_url = crossbar_api_url
+            .as_deref()
+            .unwrap_or(FALLBACK_CROSSBAR_URL);
+        let tokio_rt = Builder::new_multi_thread()
+            .thread_name("SwbPriceFetcher")
+            .worker_threads(2)
```
IliaZyrin: It looks like we only use 1 thread [here](https://github.com/mrgnlabs/eva01/pull/93/changes/36ea1a2719449324820a59e495a4c882b1980d49#diff-3f3d1e4ad129a73118e7ac00e0da3fe9f3e28584f8c19e138b0effc18d3e6ff7R164-R168), so let's just create it this way - single-threaded.

src/utils/swb_price_fetcher.rs
```diff
@@ -0,0 +1,222 @@
...
+                .filter_map(|d| d.to_string().parse::<f64>().ok())
+                .collect();
+            let conf = if valid.len() > 1 {
+                let max = valid.iter().cloned().fold(f64::NEG_INFINITY, f64::max);
+                let min = valid.iter().cloned().fold(f64::INFINITY, f64::min);
+                (max - min) / 2.0
+            } else {
+                0.0
+            };
+
```
IliaZyrin, line 240: This looks like it can be also parsed with serde, without these manual validations.

src/cli/lut_creator.rs
```diff
@@ -0,0 +1,578 @@
...
+    let total = addresses.chunks(EXTEND_CHUNK).count();
+    for (i, chunk) in addresses.chunks(EXTEND_CHUNK).enumerate() {
+        println!(
+            "[Group {}] Extending {}/{} ({} addresses)...",
+            group_num,
+            i + 1,
+            total,
+            chunk.len()
+        );
+        let out = std::process::Command::new("solana")
```
IliaZyrin: You should not need to do it using the CLI. Also no need to parse the env vars separately here, they're already part of the config.

In general, this file is hard to review/grasp. It feels like there are too many repetitive changes (for g1, g2, g3 everywhere), and too many logs and comments which can be removed. Plus the LUT creation logic is now split between this file and cache.rs (where it used to live), where only a small part left.

(comment for myself) We also decided to bring back the previous logic for specialized LUTs as a last line of defence.

src/main.rs
```diff
@@ -49,6 +49,19 @@ mod drift_idl {
 use drift_idl::*;
 
 fn main() -> Result<(), Box<dyn Error>> {
+    let args: Vec<String> = std::env::args().collect();
```
IliaZyrin: Again, this is something which should not be needed. There were no issues creating and updating LUTs using the functions in cache.rs. If there are any issues with it now, we should fix them but not split Eva into some kind of a multi-app.

src/rebalancer.rs
```diff
@@ -336,15 +343,17 @@ impl Rebalancer {
...
         );
 
-        if output_mint == Pubkey::from_str_const("So11111111111111111111111111111111111111112") {
-            return Err(anyhow::anyhow!("FIX WSOL SWAPS!"));
-        }
+        const WSOL: Pubkey = Pubkey::from_str_const("So11111111111111111111111111111111111111112");
+        // Use wrap_and_unwrap_sol=true only when wSOL is the INPUT (native SOL needs wrapping).
+        // When wSOL is the OUTPUT we want Jupiter to deposit into the wSOL ATA directly so
+        // the SPL token cache picks it up — unwrapping to native SOL would bypass that.
+        let wrap_and_unwrap_sol = input_mint == WSOL;
```
IliaZyrin: Just a reminder to fix it in a proper way, as we discussed.

src/wrappers/oracle.rs
```diff
@@ -21,39 +21,58 @@ pub trait OracleWrapperTrait {
     fn get_address(&self) -> Pubkey;
 }
 
+enum PriceSource {
```
IliaZyrin: This enum with just one value seems redundant, let's just use its inner value directly.

src/wrappers/liquidator_account.rs
```diff
@@ -277,7 +259,7 @@ impl LiquidatorAccount {
         }
 
         let liab_token_balance =
-            I80F48::from_num(self.get_token_balance_for_mint(&liab_mint).unwrap());
+            I80F48::from_num(self.get_token_balance_for_mint(&liab_mint).unwrap_or(0));
```
IliaZyrin, line 262: This should not be necessary as there is always a liability to repay, otherwise liquidation is senseless.

src/liquidator.rs
```diff
@@ -129,50 +114,34 @@ impl Liquidator {
...
-                        &mut tokens_in_shortage,
-                    ) {
+                    if let Err(e) = self
+                        .liquidator_account
+                        .liquidate(acc, &mut tokens_in_shortage)
+                    {
                         match e {
                             LiquidationError::Anyhow(e) => {
-                                error!("Failed to liquidate account {:?}", liquidatee);
+                                error!("Failed to liquidate account {:?}: {:?}", liquidatee, e);
```
IliaZyrin: It's intentional that it is not including `e` here: I just recently [removed it](https://github.com/mrgnlabs/eva01/commit/0eccacf1512b839253e2e881f785f155b6b36211#diff-d32eabcf2039db8505720a68053b90e486154a278fd69c958099a8610aae345dL150-R146), because it was cluttering the logs by duplicating what was already outputted by the program itself.

src/utils/mod.rs
```diff
@@ -234,6 +236,7 @@ pub fn log_genuine_error(prefix: &str, error: Error) {
     }
 }
 
+#[allow(dead_code)]
```
IliaZyrin: Let's just remove it if it's not used anymore.

src/utils/simulation_cache.rs
```diff
@@ -1,468 +1,22 @@
...
-                msg.contains("too large")
-            }
+            RpcError::RpcRequestError(msg) | RpcError::ForUser(msg) => msg.contains("too large"),
             _ => false,
         },
         _ => false,
     }
 }
 
+#[allow(dead_code)]
```
IliaZyrin: Let's remove all `dead_code`s from the code.

IliaZyrin (changes requested):
Great work overall!

src/utils/swb_price_fetcher.rs
k0beLeenders: I aggree this would be the right apporach if this was our own DTO, however `results` is from the swb crate so we can't control its serde annotations. We're already working with the deserialized type, we could add rust_decimal as direct dep instead of doing `to_string.parse()` to do Decimal -> f64 if you prefer that.

src/cli/lut_creator.rs
k0beLeenders: so this is a standalone script to get the 3 LUT groups which you can add to the env vars to run Eva. If we do it as part of the Eva flow we have no way of "saving" those LUTs, since the addresses are non-deterministic. We could also split this up into a separate binary to maintain LUTs if you prefer that

Regarding the previous logic for specialized LUTs, that logic was never removed. It still lives in GroupedLuts.overflow and Cache::create_targeted_lut.

k0beLeenders: a proper cleanup of this file is def in order, I agree with that

src/wrappers/liquidator_account.rs
k0beLeenders: I believe this is actually checking the liquidators token account balance not the liquidatee. If the token account doesn't exist for the liquidator (which I was running into) it needs to return 0 and correctly trigger NotEnoughFunds. If you don't add this it will fail

src/liquidator.rs
k0beLeenders: Remove this again might hide usefull logs in production, or how could you track these errors on prod?

src/utils/swb_price_fetcher.rs
IliaZyrin: Ok, it seems it would not bring much simplicity to introduce our own Normalized wrapper around it. Then yeah, let's apply Decimal -> f64, if it's easy to do and will simplify things.

src/liquidator.rs
IliaZyrin: Those will not be hidden, they will be reported even you don't print any logs in Eva at all, since it's the program which prints them.

src/utils/swb_price_fetcher.rs
k0beLeenders: we're already converting Decima -> f64 via .to_string().parse(). To use to_f64() directly we'd need to add rust_decimal as a direct dep. is that what you're suggesting, or are you happy with the current approach

src/wrappers/liquidator_account.rs
IliaZyrin: Ah, yeah, you're right. In this case it should also never happen though, because all token accounts are explicitly created on startup. So I would actually like to see it crash here, because that means there is a bug in setting up things.

src/utils/swb_price_fetcher.rs
IliaZyrin: Yes, let's add it.

IliaZyrin (approved):
Great!

PR #94, opened 2026-06-02
Iz/improvements

IliaZyrin:
This PR:
- Adds support for from_slot usage on Geyser subscriptions.
- Adds Geyser tracking for Banks.
- Removes GeyserProcessor and connects Geyser and Liquidator directly.
- Replaces the liquidation scans' "trigger" logic: now only directly affected accounts are checked for liquidatability upon detecting a Oracle/Bank/Account update from Geyser. Also, debtless accounts are ignored during scans.
- Fixes emode weight application in `compute_max_liquidatable_amounts_with_banks()`, which directly affected the amounts-to-liquidate calculation.
- Aligns the dependencies with the current program ones (Rust 1.90, Solana 3, Anchor 1.0.2).
- Adds support for breaking changes of [0.1.9](https://github.com/0dotxyz/marginfi-v2/pull/538).


IliaZyrin:
@k0beLeenders this is now ready for your review!

src/geyser.rs
```diff
@@ -213,19 +206,25 @@ impl GeyserService {
...
+                                    &mut account.data.clone().as_slice(),
+                                ) {
+                                    if marginfi_account.group != self.marginfi_group_pk {
+                                        continue;
+                                    }
+                                    self.send_update(AccountType::Marginfi, address, &account);
                                 }
 
-                                self.send_update(AccountType::Marginfi, address, &account);
+                                if let Ok(bank) =
```
k0beLeenders, line 214: It will never reach this code due to the check only checking `account_update.data.len() == MARGIN_ACCOUNT_SIZE` and not bank accounts

src/liquidator.rs
```diff
@@ -415,8 +502,17 @@ impl Liquidator {
...
             HealthPriceMode::Client(clock.clone()),
+            OnRampTransition::PreTransition,
         )?;
 
+        let emode_configs = account
+            .account
+            .lending_account
+            .get_active_balances_iter()
+            .filter(|b| !b.is_empty(BalanceSide::Liabilities))
+            .map(|b| bank_to_emode_config.remove(&b.bank_pk).unwrap());
```
k0beLeenders, line 513: bank_to_emode_config will panic if certain banks are excluded higher up

k0beLeenders (commented):
Really solid! good shit

src/liquidator.rs
IliaZyrin: It must never happen, because `bank_to_emode_config` is filled with the banks of this lending account and there cannot be any duplicates.
I could use `try_get_bank()` with `unwrap()` here, but I just don't want to create more lock contention given that we already iterated over the same set of banks in the beginning of the function.

PR #95, opened 2026-06-02
fix: underlying integration banks messing with pricing logic

k0beLeenders:


src/utils/swb_price_fetcher.rs
```diff
@@ -163,8 +177,19 @@ impl SwbPriceFetcher {
...
+            // account. For integration banks (KaminoSwitchboardPull / JuplendSwitchboardPull) the
+            // program re-applies the per-bank exchange rate on read, so writing the integration-
+            // adjusted `price_realtime` here would both double-count and clobber every other bank
+            // that shares the same feed. `source_price` is identical across all banks on a feed.
+            let raw = entry
+                .oracle_price
+                .source_price
+                .as_ref()
+                .map(|sp| &sp.price_realtime)
+                .unwrap_or(&entry.oracle_price.price_realtime);
```
IliaZyrin, line 190: So for non-integration banks `source_price` is always `None`?

k0beLeenders: correct

IliaZyrin (approved):
LGTM!

PR #97, opened 2026-06-04
fix: exclude configured mints from liquidation & repair WSOL rebalancing

k0beLeenders:


src/rebalancer.rs
```diff
@@ -181,8 +181,10 @@ impl Rebalancer {
                     }
                 }
             } else if value < min_value {
-                info!("The value of {} tokens is lower than set threshold: {} < {}. Will buy ${} worth of tokens.", mint, value.to_num::<f64>(), min_value.to_num::<f64>(), min_value.to_num::<f64>());
-                mint_to_value.insert(mint, min_value);
+                // Buy only the shortfall to reach min, not the full min_value.
```
IliaZyrin, line 184: It was intentionally buying the whole min_value to have some buffer on top of it, to not immediately having to buy it again if any liquidation spends it.

IliaZyrin (approved):
Love the simplification of only relying on the wallet and not needing a MarginFi account.

Just one small comment to discuss.

PR #99, opened 2026-06-19
feat: execution layer

k0beLeenders:
Summary

Reworks how liquidations are executed. Detection stays as-is (the vent-driven scan); execution moves behind a strategy/executor that funds each liquidation just-in-time, submits it as a Jito bundle with a sequential-RPC fallback, and only cranks oracles when a simulation says it's needed. The rebalancer collapses to a sell-only sweep so the liquidator holds a single USDC reserve.

What changed

- New execution layer (src/execution/): a LiquidationStrategy trait + Executor. The scan loop emits an intent; the executor profit-gates, assembles, simulates, and submits.
- JIT-buy funding (InventoryStrategy): buys only the shortfall of the liability ta Jupiter (inline reqwest, ExactIn — sized from the liability's oracle price,verified against the quote). No per-liability inventory.
- Jito bundle client (src/utils/jito.rs): sendBundle + status polling + simulateBundle, dynamic tip with a cap. Bundle is preferred; falls back to sequential RPC if it isn't accepted.
- Simulate-first cranking: simulate the bundle; only prepend an SWB crank on a stulate and submit only if it passes — so we don't pay to land a bundle that wouldrevert.
- Sell-only rebalancer: any non-USDC token above a small dust floor is swept back to USDC. Removed the buy/threshold/marginfi-deposit machinery.
- Removals: legacy inline liquidate() + reactive cranking + simulation_cache + th

Config

- Added (optional): JITO_BLOCK_ENGINE_URL, JITO_TIP_MAX_LAMPORTS (default 0.001 SOL), JITO_API_KEY.
- Removed: TOKEN_THRESHOLDS, DEFAULT_TOKEN_MAX_THRESHOLD, TOKEN_DUST_THRESHOLD.
- SWAP_MINT is the single reserve (USDC).

Testing

- cargo build, cargo clippy -- -D warnings, cargo test (60 pass); single solana-sdk v3 in the tree.
- Live mainnet run: clean startup → event-driven scan → sell-only sweep → JIT-buydle submit; no panics.

Known follow-ups (not in this PR)

- Detection can over-flag accounts whose SWB oracle is stale (they fail the post-crank re-sim with 6068 and are correctly skipped at no cost) — worth fixing upstream so cycles aren't spent on
phantoms, plus a short backoff for repeat phantoms.


src/execution/executor.rs
```diff
@@ -0,0 +1,470 @@
...
+struct TargetBackoff {
+    /// Consecutive assemble failures (drives the exponential cooldown).
+    failures: u32,
+    /// Don't re-attempt the target before this time.
+    retry_after: Instant,
+}
+
+/// Exponential backoff: `BASE * 2^(failures-1)`, capped at `TARGET_BACKOFF_MAX`.
+fn target_backoff_delay(failures: u32) -> Duration {
+    let shift = failures.saturating_sub(1).min(16);
```
IliaZyrin: Nitpick: 8 is more than enough
```suggestion
    let shift = failures.saturating_sub(1).min(8);
```

src/execution/executor.rs
```diff
@@ -0,0 +1,470 @@
...
+        swb_cranker: Arc<SwbCranker>,
+        signer: Keypair,
+        rpc_url: String,
+        bundle_api_key: Option<String>,
+        tip_max_lamports: u64,
+        min_profit_usd: u64,
+    ) -> Self {
+        // The Jito tip accounts are static and well-known; use the hardcoded set rather than a
+        // network fetch (the REST tip_accounts endpoint is unreliable).
+        let tip_accounts = default_tip_accounts();
```
IliaZyrin: Get rid of this member and simply use the constant slice directly (move it here).

src/execution/executor.rs
```diff
@@ -0,0 +1,470 @@
...
+    pub fn execute(
+        &self,
+        strategy: &dyn LiquidationStrategy,
+        intent: &LiquidationIntent,
+    ) -> Result<()> {
+        let liquidatee = intent.liquidatee_account.address;
+
+        // Gate on profit *before* assembling, so we don't run swap quotes or create LUTs for
+        // sub-profit targets. (Full lamport-cost-aware gating via SOL price is a follow-up.)
+        if intent.profit < self.min_profit_usd {
```
IliaZyrin: Looks like this can be moved on the caller level and the respective field can be removed from the struct.

src/execution/executor.rs
```diff
@@ -0,0 +1,470 @@
...
+                "Skipping {}: est profit ${} < min ${}",
+                liquidatee, intent.profit, self.min_profit_usd
+            );
+            return Ok(());
+        }
+
+        // Skip targets we recently failed to assemble, *before* incurring an attempt or a Jupiter
+        // quote: a liab mint with no swap route is quarantined; a target that hit a transient
+        // error (rate limit, etc.) is backed off. Both expire so recovering targets retry.
+        let liab_mint = self.intent_liab_mint(intent);
```
IliaZyrin: Should error if not found.

src/execution/executor.rs
```diff
@@ -0,0 +1,470 @@
...
+            .banks
+            .try_get_bank(&intent.liab_bank)
+            .ok()
+            .map(|b| b.bank.mint)
+    }
+
+    /// Whether a liability mint is currently quarantined (no swap route). Prunes expired entries.
+    fn is_mint_quarantined(&self, mint: &Pubkey) -> bool {
+        let now = Instant::now();
+        let Ok(mut guard) = self.quarantined_mints.lock() else {
```
IliaZyrin: It's wrong to return false here, because this error is not about the mint, but about the lock being spoiled (although should never happen anyway).

IliaZyrin: Can also be merged with `failed_targets` to decrease number of entities overall.

src/execution/executor.rs
```diff
@@ -0,0 +1,470 @@
...
+    /// Record an assemble failure: quarantine the liab mint when it has no swap route
+    /// (`NO_ROUTES_FOUND`), otherwise apply an exponential per-target backoff.
+    fn note_assemble_failure(
+        &self,
+        target: &Pubkey,
+        liab_mint: Option<Pubkey>,
+        e: &anyhow::Error,
+    ) {
+        if e.to_string().contains("NO_ROUTES_FOUND") {
+            if let Some(mint) = liab_mint {
```
IliaZyrin: Must be present

src/execution/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+pub mod inventory;
+
+use anyhow::Result;
+use solana_program::pubkey::Pubkey;
+use solana_sdk::transaction::VersionedTransaction;
+
+/// What the scan loop emits for the execution layer to act on. Today this is exactly the
+/// account the decision side already prepares; aliased here so the execution layer doesn't
+/// depend on the liquidator's internal naming.
+pub type LiquidationIntent = crate::wrappers::liquidator_account::PreparedLiquidatableAccount;
```
IliaZyrin: This can be moved to a common place to avoid introducing aliases.

src/execution/executor.rs
```diff
@@ -0,0 +1,470 @@
...
+                self.clear_target_backoff(&liquidatee);
+                plan
+            }
+            Ok(None) => {
+                debug!(
+                    "Strategy '{}' cannot handle {}; skipping",
+                    strategy.name(),
+                    liquidatee
+                );
+                return Ok(());
```
IliaZyrin, line 185: Better name the function `try_execute()` then.

src/execution/executor.rs
```diff
@@ -0,0 +1,470 @@
...
+        };
+
+        // Simulate-first crank detection. The outcome dispatches four ways:
+        //  - ran & ok                 -> bundle send, no crank
+        //  - ran & stale (0x17a1)     -> prepend crank, bundle send
+        //  - ran & other prog error   -> skip (doomed on-chain; don't burn tip+fees)
+        //  - couldn't run (infra err) -> crank all feeds + sequential RPC send (no bundle)
+        // Temp LUTs created during assembly are always cleaned up afterwards, whatever the path.
+        let temp_luts = plan.temp_luts;
+        let mut txs = plan.txs;
```
IliaZyrin: Nitpick: can be matched in one-line.

src/execution/executor.rs
```diff
@@ -0,0 +1,470 @@
...
+                target,
+                delay.as_secs(),
+                entry.failures
+            );
+        }
+    }
+
+    /// Deactivate any temporary LUTs created during assembly (best-effort; logs on failure), and
+    /// close earlier deactivated LUTs whose cooldown has elapsed to reclaim their rent.
+    fn deactivate_temp_luts(&self, temp_luts: &[Pubkey]) {
```
IliaZyrin: Can probably be moved to a separate thread.

src/execution/executor.rs
```diff
@@ -0,0 +1,470 @@
...
+        {
+            warn!("Failed to deactivate temporary LUT(s) {temp_luts:?}: {e}");
+        }
+    }
+
+    /// Build a crank tx for the intent's stale feeds, unless they were all cranked very recently.
+    fn build_crank_if_needed(
+        &self,
+        intent: &LiquidationIntent,
+    ) -> Result<Option<VersionedTransaction>> {
```
IliaZyrin: Can probably be replaced with just `Option<VersionedTransaction>`

src/execution/executor.rs
```diff
@@ -0,0 +1,470 @@
...
+        }
+        // Pick a tip account pseudo-randomly to spread tip load across them (Jito's guidance).
+        let idx = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .map(|d| d.subsec_nanos() as usize)
+            .unwrap_or(0)
+            % self.tip_accounts.len();
+        let tip_account = &self.tip_accounts[idx];
+        let tip_lamports = self.tip_estimator.current_tip();
+        let ix = transfer(&self.signer.pubkey(), tip_account, tip_lamports);
```
IliaZyrin: This whole logic for creating a tip ix can be moved out of executor, e.g. to `tip_estimator`.

src/execution/executor.rs
```diff
@@ -0,0 +1,470 @@
...
+                )
+                .map_err(|e| {
+                    anyhow!(
+                        "Sequential send failed at tx {}/{}: {}",
+                        i + 1,
+                        txs.len(),
+                        e
+                    )
+                })?;
+            info!("Sequential tx {}/{} confirmed: {}", i + 1, txs.len(), sig);
```
IliaZyrin, line 586: Just a general thought: should probably track how often this path gets used. Depending on the results we might want to optimize it.

src/execution/inventory.rs
```diff
@@ -0,0 +1,311 @@
...
+        v.get("swapTransaction")
+            .and_then(|s| s.as_str())
+            .map(|s| s.to_string())
+            .ok_or_else(|| anyhow!("Jupiter swap response missing swapTransaction: {v}"))
+    }
+}
+
+/// Amount of the liability token to buy: the part of `repay_amount` not already in the wallet.
+/// Uses the same `u64` truncation as the repay ix so the buy lands exactly what repay consumes.
+fn buy_shortfall(repay_amount: I80F48, wallet_balance: u64) -> u64 {
```
IliaZyrin: Most of this can possibly be moved to super-dex-agg.

src/utils/jito.rs
```diff
@@ -0,0 +1,527 @@
...
+    }
+
+    fn fetch_tip_lamports(&self) -> Result<u64> {
+        let resp: Value = self.http.get(TIP_FLOOR_URL).send()?.json()?;
+        parse_tip_floor_lamports(&resp)
+    }
+}
+
+/// Parse `ema_landed_tips_50th_percentile` (SOL) from the tip_floor response into lamports.
+fn parse_tip_floor_lamports(resp: &Value) -> Result<u64> {
```
IliaZyrin, line 401: This can be replaced with serde I guess.

src/utils/swb_cranker.rs
```diff
@@ -84,98 +64,24 @@ impl SwbCranker {
...
-    }
-
     fn build_crank_transaction(&self, swb_oracles: Vec<Pubkey>) -> Result<VersionedTransaction> {
         self.tokio_rt
             .block_on(self.build_crank_transaction_async(swb_oracles))
     }
 
+    /// Build a standalone SWB crank tx for the given feeds, for the execution layer to prepend to
+    /// a Jito bundle (simulate-first crank). Public wrapper over `build_crank_transaction`.
+    pub fn build_crank_tx(&self, swb_oracles: Vec<Pubkey>) -> Result<VersionedTransaction> {
```
IliaZyrin: I suggest just making `build_crank_transaction` `pub`.

src/wrappers/liquidator_account.rs
```diff
@@ -200,159 +159,82 @@ impl LiquidatorAccount {
...
+    /// Assemble (compile + sign) the liquidation transaction for the given account and amounts.
+    /// Pure tx construction used by the execution-layer `InventoryStrategy`. Returns the signed tx,
+    /// its instructions (for the LUT-retry path), and an optional temporary-LUT key created to fit
+    /// an oversized tx (the caller must deactivate it once the tx lands).
+    pub fn build_liquidate_tx(
         &self,
-        account: PreparedLiquidatableAccount,
-        tokens_in_shortage: &mut HashSet<Pubkey>,
-    ) -> Result<(), LiquidationError> {
-        let PreparedLiquidatableAccount {
```
IliaZyrin, line 208: I would keep this decomposition.

src/wrappers/liquidator_account.rs
```diff
@@ -483,285 +354,41 @@ impl LiquidatorAccount {
...
+                .flat_map(|ix| ix.accounts.iter().map(|a| a.pubkey))
+                .collect::<HashSet<_>>()
+                .into_iter()
+                .collect();
+            let lut = self
+                .cache
+                .get_targeted_lut(&self.rpc_client, &self.signer, all_accounts)?;
+            temp_lut = Some(lut.key);
+
+            let mut fitted_luts = luts;
```
IliaZyrin: Can just make `luts` itself `mut`.

src/wrappers/liquidator_account.rs
```diff
@@ -483,285 +354,41 @@ impl LiquidatorAccount {
...
-                    ..Default::default()
-                },
-            );
-        debug!(
-            "Depositing result for account {:?} (without preflight check): {:?} ",
-            marginfi_account, res
-        );
 
-        Ok(())
+        Ok((tx, ixs, temp_lut))
```
IliaZyrin: We probably should just return `ixs` and `temp_lut` here, to have the final tx building logic in a single place only.

src/config.rs
```diff
@@ -27,13 +20,16 @@ pub struct Eva01Config {
...
     pub slippage_bps: u16,
-    pub token_thresholds: HashMap<Pubkey, TokenThresholds>,
-    pub default_token_max_threshold: I80F48,
-    pub token_dust_threshold: I80F48,
     pub titan_ws_endpoint: String,
     pub titan_api_key: String,
     pub jupiter_api_key: String,
     pub use_fumarole: bool,
+    /// Jito block-engine `sendBundle` endpoint; `None` uses the executor's built-in default.
+    pub jito_block_engine_url: Option<String>,
```
IliaZyrin, line 28: Do we actually use this one?

src/liquidator.rs
```diff
@@ -480,7 +488,7 @@ impl Liquidator {
         let all = asset_weight_maint - liab_weight_maint * liquidation_discount;
 
         if all >= I80F48::ZERO {
-            warn!("Account {:?} has no liquidatable amount: {:?}, asset_weight_maint: {:?}, liab_weight_maint: {:?}", account.address, all, asset_weight_maint, liab_weight_maint);
```
IliaZyrin, line 483: this was intentionally `warn` afair, because it's a pretty rare case, worth investigating when happens.

IliaZyrin (approved):
Awesome improvements, nice job!

